PR #34, opened 2020-11-03
Feature/reorganize repo

QuentinGruber:


QuentinGruber:
/out wasn't excluded  ^^ is why DeepScan isn't happy rn

PR #64, opened 2020-12-01
Feature/preventSecurityIssueHaxRun

### LINKED ISSUE #58: we need to prevent security issue with /hax run

QuentinGruber:
Close #58 

what do you thing of that @RhettVX ?

have you found an higher value than 15.0 that is working ?


QuentinGruber:
ok since people are having fun with that i will revert this and only didn't allow the values that make the game crash

PR #130, opened 2021-03-13
feature/MongoZone

QuentinGruber:
todo:

- [x] force reload mongo stuff

- [x] check that it did not break anything for the solo play


src/servers/LoginServer/loginserver.ts
```diff
@@ -104,6 +95,9 @@ export class LoginServer extends EventEmitter {
               };
               data = this._protocol.pack("LoginReply", falsified_data);
               this._soeServer.sendAppData(client, data, true);
+              if(false && !this._soloMode){ // disable this for now ( WIP )
```
QuentinGruber, line 98: finish that before merge

PR #145, opened 2021-04-25
Release/0.8.14

QuentinGruber:
- Add `/hax spamPoliceCar`
- Spawn doors in the world ( not usable tho and 20% of the doors are kinda glitched in walls but when they will be usables doors it will be fixed )
- Use `/hax despawnObjects` to despawn all doors in the world and they won't respawn until you restart your server
- Remove zombie character since we can use `/hax changeModel 9001` to do that properly now


src/packets/ClientProtocol/ClientProtocol_860/h1z1packets.js
```diff
@@ -5191,7 +5191,23 @@ var packets = [
...
+      fields: [
+        { name: "unknown1", type: "byte", defaultValue: 0 },
+        { name: "unknown2", type: "byte", defaultValue: 0 },
+        {
+          name: "characterId",
+          type: "uint64",
+          defaultValue: "0x0000000000000000",
+        },
+        { name: "modelId", type: "uint32", defaultValue: 0 },
+        { name: "unknown3", type: "uint32", defaultValue: 0 },
```
QuentinGruber, line 5207: maybe first person model ?

PR #208, opened 2021-05-29
Rotation for spawnNpcModel

avcio9:
thx to GuinnesRules for using his code to fix rotations


QuentinGruber:
eul2quat is already imported in h1emu ( base your work on /dev not /master )

what's the point of this ? like character rotation is already saved in the format we want

QuentinGruber (approved):

PR #212, opened 2021-05-30
Update dynamicWeather.ts

avcio9:
Added wind to system, it gets stronger in autumn/winter.
Added clouds accumulating more during autumn/winter, there is a chance for small clouds during summer.
Rain will now slowly grow stronger and later weaken. I thought it was a 0/1 toggle because its already strong at 1, but it can get stronger :P.
Sun will now change axis as smooth as it can - sadly game doesnt accept floats so it changes by 1 every 3 ticks


QuentinGruber (approved):
noice :P

PR #241, opened 2021-06-14
Engine off after dismount

avcio9:


src/types/zoneserver.ts
```diff
@@ -1,6 +1,7 @@
 export interface Client {
   isLoading: boolean;
   isMounted: boolean;
+  selectedVehicle: string;
```
QuentinGruber, line 4: we should do that this field can be empty ( with ? )  and to delete it rather that setting it to "0"

QuentinGruber (changes requested):
also selectedVehicle isn't a great name and we could merge isMounted and the new one, there is no need to have 2 things that say the same thing.

QuentinGruber (approved):

PR #249, opened 2021-06-16
Adjusted sprint speed & removed unneccesary flags

avcio9:
Right now other flags seems unnecesary, perhaps they will be needed for something in the future but right now the movement seems better without them.


avcio9:
Halted for now

PR #252, opened 2021-06-17
/hax nameChange and password protect

avcio9:


QuentinGruber:
to be honest this kind of change should be on our custom implementation of the server not on the library

PR #265, opened 2021-06-23
Vehicle destroy and damage levels, hax siren and headlights

avcio9:
Its for singleplayer only, so i based the "health" on adding value to a variable and didnt bother about seperate health for multiple vehicles. If we apply siren or headlights then we cant turn them off, for that we need a proper effect packet, as well as for it to work in multiplayer.
I based my work on master branch cause i had problems with /hax drive on dev

### DIFF
--- a/src/servers/ZoneServer/commands/hax.ts
+++ b/src/servers/ZoneServer/commands/hax.ts
@@ -7,8 +7,53 @@ const debug = require("debug")("zonepacketHandlers");
 
 let isSonic = false;
 let isVehicle = false;
+let headlightType = 0;
 
 const hax: any = {
+  siren: function (server: ZoneServer, client: Client, args: any[]) {
+    if (client.isPolice) {
+      server.sendData(client, "Mount.DismountResponse", {
+        characterId: client.character.characterId,
+      });
+      server.sendData(client, "Mount.MountResponse", {
+        characterId: client.character.characterId,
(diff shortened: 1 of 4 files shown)


src/servers/ZoneServer/zonepackethandlers.ts
```diff
@@ -1586,6 +1691,22 @@ const packetHandlers: any = {
...
+      switch (vehicleTypeId) {
+        case 1:
+          client.mountedVehicleType = "offroader";
+          break;
+        case 2:
+          client.mountedVehicleType = "pickup";
+          break;
+        case 3:
+          client.mountedVehicleType = "policecar";
+          client.isPolice = true;
```
QuentinGruber, line 1704: it's kinda a duplication ? like if you wanna check if "isPolice" you can already check like that `if client.mountedVehicleType === policecar`

src/servers/ZoneServer/commands/hax.ts
```diff
@@ -7,8 +7,53 @@ const debug = require("debug")("zonepacketHandlers");
 
 let isSonic = false;
 let isVehicle = false;
+let headlightType = 0;
```
QuentinGruber, line 10: define it before the switch case

QuentinGruber (changes requested):

PR #273, opened 2021-06-23
fix full vehicle, hax, add prop spawn and interactions

avcio9:
vehicle damage and destroy worked well already, still have to fix interactionstring on mounted cars spawned by server.

Some props dont send interactionstrings or you have to find a weird angle to trigger it (they may be covered by client-spawned objects) 
We have to find a way to despawn the default spawned objects


QuentinGruber (approved):

QuentinGruber:
@avcio9 About despawning, there is a packet called despawnNpc if i remember well but no idea on how to get the id of the client side spawned object

src/servers/ZoneServer/zoneserver.ts
```diff
@@ -536,7 +546,7 @@ export class ZoneServer extends EventEmitter {
     ];
     this._dummySelf = require("../../../data/2015/sampleData/sendself.json"); // dummy this._dummySelf
     if (String(client.character.characterId) === "0x0000000000000001") {
-      // for fun 🤠
+      // for fun ­Ъца
```
QuentinGruber, line 549: What does it mean ? 😀

src/servers/ZoneServer/zoneserver.ts
```diff
@@ -795,7 +806,7 @@ export class ZoneServer extends EventEmitter {
 
       this.sendData(
         client,
-        "PlayerUpdate.RemovePlayerGracefully",
```
QuentinGruber, line 798: Why changing that? With this packet the entity is still in the game i don't think we should use it

QuentinGruber (changes requested):

src/servers/ZoneServer/zoneserver.ts
avcio9: Pretty :P

src/servers/ZoneServer/zoneserver.ts
avcio9: RemovePlayer packet for less resource usage, if it doesnt work that way ima change it back

avcio9:
> @avcio9 About despawning, there is a packet called despawnNpc if i remember well but no idea on how to get the id of the client side spawned object

All the .zone objects have unique id's in json files, maybe thats the way

QuentinGruber (approved):

PR #288, opened 2021-07-01
Fix props

avcio9:
Updated z1_Props with more objects, fixed rotations (some of them need eul2quat)
Added some basic interactions for props


src/servers/ZoneServer/workers/createBaseEntities.ts
```diff
@@ -771,17 +781,47 @@ function createProps(server: ZoneServer) {
       MODEL_FILE_NAME: propType.actorDefinition,
     })?.ID;
     propType.instances.forEach((propInstance: any) => {
+      let rotationfix = propInstance.rotation;
+      if (
+        propType.actorDefinition.includes(
```
QuentinGruber, line 786: this is pretty ugly, maybe we can find a better way

src/servers/ZoneServer/zonepackethandlers.ts
```diff
@@ -1793,7 +1810,8 @@ const packetHandlers: any = {
         propToSearch.position
       )
     ) {
-      let interactType = "0";
+      let interactType;
+      let logoutTime = 0;
```
QuentinGruber, line 1814: maybe interactionTime ?

QuentinGruber (changes requested):
Nice job btw :P some small things to maybe fix, i will do these this night if you don't have time.

src/servers/ZoneServer/zonepackethandlers.ts
avcio9: Im gonna use the same as logoutTimer, just ranamde it to "timer" since its the same for everything and there is no need to set it up twice. about createBase entities, i've included eul2quat for some actors in json file so there is no need to do it in script

avcio9:
Spawning all entities at first loading and not removing them by distance fixes server packet overload error
It causes a few secs of lag when they first load, but compensates less lag later. Some people may have lowered fps in high demand locations like PV until we figure how to remove default props

QuentinGruber (approved):

PR #319, opened 2021-07-11
Collision damage

avcio9:
Not sure if collision.damage packet is 100% correct, it sends values each time you touch something so i'm using falling speed for it aswell
Resource value wasnt saving after picking up stuff so i changed it in playerselect (it was just changing constants)


src/types/zoneserver.ts
```diff
@@ -3,6 +3,7 @@ export interface Client {
   firstLoading: boolean;
   isLoading: boolean;
   isInteracting: boolean;
+  falling: number;
```
QuentinGruber, line 6: since it's linked to character or vehicle, i would move that to client.character or vehicle

QuentinGruber (changes requested):

QuentinGruber (dismissed):

QuentinGruber (approved):

PR #389, opened 2021-08-15
Use functions for event hooks

Z1Meme:
Needed so 2016 server is able to override hooks on the 2015 server (ex. onZoneDataEvent())


QuentinGruber (approved):
nice, sorry that i didn't do that earlier 😉

PR #458, opened 2021-09-15
added /hax spectate command

Z1Meme:
ye


QuentinGruber (approved):
nice i've try to do that once but doesn't thought about using it as a vehicle :P

PR #465, opened 2021-09-16
feature/MultiplayerVehicleMovement

avcio9:
Basic movement is working, tho we need some more packets to make the tires move etc


src/packets/ClientProtocol/ClientProtocol_860/h1z1packets.ts
```diff
@@ -168,31 +168,31 @@ function readPositionUpdateData(data: Buffer, offset: number) {
 
   if (obj.flags & 0x40) {
     var v = readSignedIntWith2bitLengthValue(data, offset);
-    obj["frontTilt"] = v.value / 100; // not 100% sure about name
+    obj["unknown7_float"] = v.value / 100; // not 100% sure about name
```
QuentinGruber, line 171: are the var names like frontTilt were the right ones ? if yes we should change "unknown7_float" on the h1z1protocol instead of renaming this here

PR #473, opened 2021-09-19
feature/Open&Close door & positionUpdate var names

avcio9:
Not sure if thats 100% how daybrak was doing it
After using doors for too long their collidable objects misplaces visual ones
As a workaround for now they are removed and respawned after 6 open/close loops
Setting isVehicle for npc breaks interactionstrings and collision, somehow adding g:127 to color fields restores it
Some doors are open by default, that makes them open into walls or stuff like that, i dont know how to identify them atm.
Same with double doors, some of them will open in wrong directions


QuentinGruber:
> Not sure if thats 100% how daybrak was doing it
> After using doors for too long their collidable objects misplaces visual ones
> As a workaround for now they are removed and respawned after 6 open/close loops
> Setting isVehicle for npc breaks interactionstrings and collision, somehow adding g:127 to color fields restores it
> Some doors are open by default, that makes them open into walls or stuff like that, i dont know how to identify them atm.
> Same with double doors, some of them will open in wrong directions

About doors in wrong direction i will fix that on this PR before merging that was intentionnal

PR #487, opened 2021-09-25
Feature/inputAndOutputThreading

QuentinGruber:
This is not ready at all.

This need some more love & tunning.

+ I need to make proper benchmark to compare server's performance with and without this feature

todos:

- [ ] Some threads aren't killed after client logout, need to fix that


QuentinGruber:
hm based on benchmark , this is 10 times slower that current /dev :(

QuentinGruber:
reopen that, i need to try it back with more threads allocated to libuv

PR #502, opened 2021-10-14
h1emu protocol

Z1Meme:
…r h1emuserver; udp connection established between login and zone server (localhost only); test packet sent from login server to zone and vice versa


QuentinGruber:
still need to cover packet loss

QuentinGruber (approved):

PR #514, opened 2021-10-27
Character death/respawn

avcio9:
needs #515 merged to work

Player deaths working correctly, workaround for respawn but works nice
Vehicles now have 10000 health and take proper damage on collision.
Added /hax godmode (sonic was a nightmare 😄 )

Also we need an item-respawning system before using this


src/servers/ZoneServer/commands/hax.ts
```diff
@@ -1013,6 +1013,12 @@ const hax: any = {
     });
     server.sendChatText(client, "Back to normal size");
   },
+  godmode: function (server: ZoneServer, client: Client, args: any[]) {
+    let godmode = "";
+    client.character.godMode = client.character.godMode ? false : true;
```
QuentinGruber, line 1018: This ternary is useless since it's already a boolean

QuentinGruber (approved):
Why should we wait to have item respawn? It can be in another PR 🤠

QuentinGruber:
`/hax drive` make the character die :(

avcio9:
> `/hax drive` make the character die :(

Its caused by collision packet sent by client with great damage. You should use godmode while spawning vehicles, or we could assign it temporarily to players spawning a vehicle

QuentinGruber:
> > `/hax drive` make the character die :(
> 
> Its caused by collision packet sent by client with great damage. You should use godmode while spawning vehicles, or we could assign it temporarily to players spawning a vehicle

I think we should assign temp invulnerability for commands like that

QuentinGruber (approved):

QuentinGruber:
#517 should be done before this being merge in /master

PR #526, opened 2021-11-05
Added packets

thegrreat1:
Vehicle.HealthUpdateOwner set's the vehicle health in OwnedVehicle.Info datasource.
Vehicle.AccessType locks the vehicle when l is pressed. Now we just need to add a server variable to check if the vehicle is locked :)


QuentinGruber (approved):
🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗 🚗

PR #544, opened 2021-11-11
Added PlayerUpdate.SlotCompositeEffectOverride

thegrreat1:
Let's us change the effect of the current slot.. Since we cant change weapon using 1, 2 ,3 4,  1 is selected by default. So only works on slot1 at the moment :)


QuentinGruber (approved):
🚀

PR #546, opened 2021-11-11
fix: synchronization & no player movement

avcio9:
Doing it this way server can be up to almost 50 days without the gametime value going out of uint32 range


QuentinGruber (approved):
🥇 🥇 🥇 🥇 🥇 🥇

PR #602, opened 2021-11-23
Reversed packets and added /dev hideme

thegrreat1:
✖️ Features in this update ✖️
----------------------------------------------------------------------------------------------------------------------------------------
- Player will now bleed when taking 40% damage in one hit or health is less than 20% and taking any damage.
  Could change the value if wanted!
- Added /hax bandage. Will increase health slowly and stop bleeding when health is above 20%.

- Added /dev hideme "So @QuentinGruber  can hide himself from all the fans on live server 🤪 ".
- Decrease stamina if character is bleeding and running. And disable it from increasing while character is bleeding.

- Interactable burlapbag now spawns ontop of player on death
- --------------------------------------------------------------------------------------------------------------------------------------

### DIFF
--- a/src/packets/ClientProtocol/ClientProtocol_860/h1z1packets.ts
+++ b/src/packets/ClientProtocol/ClientProtocol_860/h1z1packets.ts
@@ -6212,7 +6212,15 @@ var packets = [
           ]
       }
   ],
-  ["ClientUpdate.CollectionRemoveEntry", 0x110900, {}],
+  ["ClientUpdate.CollectionRemoveEntry", 
+    0x110900,
+    {
+        fields: [
+              { name: "unknownDword1", type: "uint32", defaultValue: 0 },
+              { name: "unknownDword2", type: "uint32", defaultValue: 0 },
+          ]
+      }
+  ],
   [
     "ClientUpdate.UpdateLocation",
     0x110a00,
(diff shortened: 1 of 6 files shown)


thegrreat1:
*NOTICE TO MYSELF* Players zoning in after character has done /dev hidden can still see the hidden player.. Need to add a check for that.

thegrreat1:
> 

Solved!

thegrreat1:
> _NOTICE TO MYSELF_ Players zoning in after character has done /dev hideme can still see the hidden player.. Need to add a check for that. "Dunno how to do it yet tho 'sadface'"..

Solved!

thegrreat1:
Should we add the chance of bleeding when taking 40% damage or should i delete it? You decide @QuentinGruber :)

Bleeding can be stopped by using /hax bandage.

QuentinGruber:
> Should we add the chance of bleeding when taking 40%+ damage or should i delete it? You decide @QuentinGruber :)
> 
> Bleeding can be stopped by using /hax bandage.

I think it's a good idea 😛

thegrreat1:
Everything seems to be working so im done with this one :octocat:

thegrreat1:
Sorry for the spaghetti code... Everything seems to be working fine now :)

QuentinGruber (approved):

PR #606, opened 2021-11-25
Feature/DTO's, Bugfix/vehicles dropping underground on TP

avcio9:
Default zone props are now deleted alongside DTO's
DTO's will respawn 30min after being destroyed
There is a radius check for DTO's respawning that will prevent it if there are vehicles nearby
Gave up on using positionUpdate for signs (it looked like $2 cheap mod anyway).
Gonna rework signs when we get rangdoll working properly.
Fixed vehicles getting under ground,
Vehicles will now take damage when upside-down


src/servers/ZoneServer/zoneserver.ts
```diff
@@ -1336,6 +1351,42 @@ export class ZoneServer extends EventEmitter {
     }
   }
 
+DTOhit(client: Client, packet: any) {
+    if (
+      packet.data.objectCharacterId != client.vehicle.mountedVehicle &&
```
QuentinGruber, line 1356: Why necessary checking that's the vehicle mounted by the client that send this packet?

src/servers/ZoneServer/zoneserver.ts
```diff
@@ -1931,6 +1982,70 @@ export class ZoneServer extends EventEmitter {
   spawnProps(client: Client): void {
     this.spawnNpcCollection(client, this._props);
   }
+  
+  customizeDTO(client: Client): void {
+    const DTOArray: any = [];
```
QuentinGruber, line 1987: i saw that "any" is used a lot maybe we could make a type / class for dto ?

src/servers/ZoneServer/zoneserver.ts
```diff
@@ -1931,6 +1982,70 @@ export class ZoneServer extends EventEmitter {
...
+      ) {
+        client.npcsToSpawn.push(DTOObject);
+        client.spawnedDTOs.push(DTOObject);
+      }
+    }
+  }
+
+  respawnDTOs(): void {
+    for (const DTO in this._destroyableTimestamp) {
+      const DTOObject = this._destroyableTimestamp[DTO];
```
QuentinGruber, line 2030: beside having a void running on the main thread and checking every worldroutine / serverTicks we should use timeouts for this

QuentinGruber (changes requested):

src/servers/ZoneServer/zoneserver.ts
avcio9: If you hit a non dto object Like building both characterId's belong to your vehicle, it prevents IT from being processed further

src/servers/ZoneServer/zoneserver.ts
QuentinGruber: I will do that in another PR, we need classes for npcs also and the dto class will inherit from it

src/servers/ZoneServer/zoneserver.ts
avcio9: done

QuentinGruber (approved):

PR #611, opened 2021-11-28
:star2: Bleeding prototype a little bit better

thegrreat1:


thegrreat1:
For some reasons the burlap bag will still colide with the player ragdoll but aint much i can do about it right now :)

QuentinGruber (approved):

PR #669, opened 2022-01-23
Feature/useNodePrefixForInternalModules

QuentinGruber:


QuentinGruber:
can't be used until we drop support for nodejs14

PR #732, opened 2022-03-20
cool stuff idk

Z1Meme:


src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -3245,6 +3309,96 @@ export class ZoneServer2016 extends ZoneServer2015 {
...
+      delete client.character.timeouts[a];
+    }
+  }
+
+  applyMovementModifier(client: Client, modifier: number, type: string) {
+    this.multiplyMovementModifier(client, modifier);
+    switch (type) {
+      case "wellRested":
+        if (client.character.timeouts["wellRested"]) {
+          client.character.timeouts["wellRested"]._onTimeout();
```
QuentinGruber, line 3325: Are you sure it does what you want ? ._ontimeout() doesn't "destroy" the timer right ? so maybe clear it on this case "clearTimeout" instead of an if statement in the timer callback

QuentinGruber (changes requested):

src/servers/ZoneServer2016/zoneserver.ts
avcio9: i tried to do it that way, somehow couldnt stop it from executing, thats why it checks if the timer still exists when it executes and returns the func if false

avcio9: I do it like that to execute it early, the movement modifier sucks, you cant just set it to 1 to be like before. if you multiplied the speed by 1.2, then you have to divide it by  1/1.2 to get it back to normal

QuentinGruber (approved):

PR #735, opened 2022-03-25
Finish client container logic

Z1Meme:


Z1Meme:
fully tested and ready, however still can't move items into loadout slots (reverse issue)

src/servers/ZoneServer2016/commands/hax.ts
```diff
@@ -644,12 +644,14 @@ const hax: any = {
       server.updateEquipmentSlot(client, 3);
     }
   },
+  /*
   addallitems: function (server: ZoneServer2016, client: Client, args: any[]) {
```
QuentinGruber, line 648: so we disable this command ?

src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -278,9 +273,9 @@ export class zonePacketHandlers {
       client: Client,
       packet: any
     ) {
-      const characterId = packet.data.characterId;
-      const damage = packet.data.damage;
-      const vehicle = server._vehicles[characterId];
+      const characterId = packet.data.characterId,
+      damage = packet.data.damage,
```
QuentinGruber, line 277: i've to admit that i'm not a huge fan of this syntax, i will just do a quick bench to be sure if it doesn't change anything ( it shouldn't )

QuentinGruber (approved):
i will test this in-game and also run a benchmark to see if a particular function slow something :)

src/servers/ZoneServer2016/zonepackethandlers.ts
QuentinGruber: yep it doesn't change anything the only bad thing is  that if you mess up a comma like that the variable become a global one ( which is never cleaned from memory) but as long at it never happen it's good 👍

src/servers/ZoneServer2016/classes/character.ts
```diff
@@ -106,7 +106,7 @@ export class Character2016 extends Character {
         const { stamina, food, water, virus, health, bleeding } =
           client.character.resources;
         const { isRunning } = client.character;
-        if (isRunning) {
+        if (isRunning && (client.vehicle.mountedVehicle == "" || !client.vehicle.mountedVehicle)) {
```
QuentinGruber, line 109: i think it's not the good way to do it, like isRunning whould be equal to false when driving a vehicle

QuentinGruber (changes requested):

src/servers/ZoneServer2016/classes/character.ts
avcio9: Setting isRunning to false when entering vehicle doesnt work 100% of the time, sometimes server received position packet from client after mounting so it gets set back to true

QuentinGruber (approved):

PR #787, opened 2022-04-04
Wrong opcode "Construction.Unknown"

sofresh007:
From what i can see we have wrong opcode for this unknown packet and we had 2x 400 packets and the unknown packet supposed to have opcode x500


QuentinGruber:
> From what i can see we have wrong opcode for this unknown packet and we had 2x 400 packets and the unknown packet supposed to have opcode x500

Hey, i'm very happy to see you contributing :) Could you prove what you are saying? For example the address where you found this information?

Z1Meme:
> > From what i can see we have wrong opcode for this unknown packet and we had 2x 400 packets and the unknown packet supposed to have opcode x500
> 
> Hey, i'm very happy to see you contributing :) Could you prove what you are saying? For example the address where you found this information?

He is correct
`
switch ( *v6 )
      {
        case 2:
          HandlePlacementResponsePacket(a1, buffer, bufferSize_);
          break;
        case 4:
          HandlePlacementFinalizeResponse(a1, buffer, bufferSize_);
          break;
        case 5:
          HandleUnknownContructionPacket(a1, buffer, bufferSize_);
          break;
      }
`

Z1Meme (dismissed):

Z1Meme (approved):

Z1Meme:
@QuentinGruber not sure why "Labeler" is not successful, does it matter?

QuentinGruber:
> @QuentinGruber not sure why "Labeler" is not successful, does it matter?

it's just an issue with chris not being a validated contributor, it will be fixed

PR #840, opened 2022-04-30
Feature/ClientPath

QuentinGruber:
- [ ] Find when a ClientPath.Request is sent by the client and trigger it


QuentinGruber:
For me it was used in everquest to make your player move where you click that's all

PR #903, opened 2022-06-01
Update ui.ts

sofresh007:
Added Ui.HeadShotFeedback packet struct.


QuentinGruber:
Hi, plz resend a PR with only the changes on src/packets/ClientProtocol/ClientProtocol_1080/ui.ts

PR #941, opened 2022-06-15
Feature/unbufferedData

QuentinGruber:


QuentinGruber:
This will be useful for weapon shooting @Z1Meme , using the unbuffered variant of a sendData function will send the data as soon as possible without trying to buffer the data with other packets.

Issue #954, opened 2022-06-21
type error


QuentinGruber:
AddLightweightPc : TypeError: Cannot read properties of undefined (reading 'length')
Could not pack data schema for AddLightweightPc

PR #956, opened 2022-06-21
Feature/dataschemaCleaningAndOptimizations

QuentinGruber:


QuentinGruber:
Not really fixing #954 but with the added log it would be easy to fix it when this go live ( and creating a better issue )

PR #964, opened 2022-06-29
Feature/Props

avcio9:


QuentinGruber (changes requested):

Z1Meme:
@avcio9 Any lootable props need to be FullNpcs since they use loadout for the container

PR #975, opened 2022-07-13
Feature/Placement

avcio9:


QuentinGruber:
Shouldn't we merge this ? @Z1Meme @avcio9. Since it's already used on community server.

avcio9:
there is an issue with with simpleNpcs constructions, their rotation gets tweaked sometimes, also at this point we cant have hp bars to show up using lightWeightNpcs, so for now we can either have hp bars on bases or proper rotation. 
Pr still needs a little more work,

.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,7 @@
+{
+  "ExpandedNodes": [
```
QuentinGruber: we need to add .vs/ to .gitignore

Z1Meme (approved):
👍

PR #1030, opened 2022-07-25
World Saves

Z1Meme:


Z1Meme:
@QuentinGruber Character saving works for both single and multiplayer (should not require a database reset), and vehicle saving is ready for when we have positionupdates fixed since we have some cleanup to do there and they don't work correctly rn.

Z1Meme:
Branch can be deleted btw

QuentinGruber:
I'm reviewing that today @Z1Meme 🤠

QuentinGruber:
Also this should be a feature easy to disable ( some servers don't want to store characters data and others )

QuentinGruber:
So i've spot 3 issues:

* Old characters can't be used
* Death from collision damage on spawn happens almost everytime
* I saw the log once and it only concern solo play but we shouldn't save circular structure to json ( didn't know which structure it was )

For the code and the way you structured it, it's well done 🤠

src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -814,7 +814,7 @@ export class zonePacketHandlers {
         server.setGodMode(client, false);
         client.character.tempGodMode = false;
       }
-      client.character.positionUpdate = packet.data;
+      //client.character.positionUpdate = packet.data;
```
QuentinGruber, line 817: if we remove it, we should remove the usage of it when sending LightweightToFullPc

QuentinGruber (changes requested):

QuentinGruber:
> So i've spot 3 issues:
> 
> * Old characters can't be used
> * Death from collision damage on spawn happens almost everytime
> * I saw the log once and it only concern solo play but we shouldn't save circular structure to json ( didn't know which structure it was )
> 
> For the code and the way you structured it, it's well done 🤠

Also 2015 version is broken with these changes, maybe we should create legacy functions for it

Z1Meme:
> > So i've spot 3 issues:
> > 
> > * Old characters can't be used
> > * Death from collision damage on spawn happens almost everytime
> > * I saw the log once and it only concern solo play but we shouldn't save circular structure to json ( didn't know which structure it was )
> > 
> > For the code and the way you structured it, it's well done 🤠
> 
> Also 2015 version is broken with these changes, maybe we should create legacy functions for it

- Old characters should just be deleted, they don't really matter anyways
- Death / damage from collision damage is fixed
- Can't reproduce that circular json issue
- Can't reproduce the 2015 issue, it works for me on both singleplayer and multiplayer, and as far as I know nothing was edited having to do with 2015

Z1Meme:
> Also this should be a feature easy to disable ( some servers don't want to store characters data and others )

Done
`process.env.ENABLE_SAVES`

Z1Meme:
@QuentinGruber this is ready to merge when you have some time

QuentinGruber:
I was thinking we could add some versioning to this. In the Worlds collection each world should have the version of the save system it generated, so that it wipes the world automatically if a newer version is used

QuentinGruber (approved):
🥳

PR #1037, opened 2022-08-02
Basic farming implementation

655456:
Added a basic implementation of farming with placement


Z1Meme:
@655456 What is your discord? This looks great so far but you should refer to https://github.com/QuentinGruber/h1z1-server/tree/master/src/servers/ZoneServer2016/classes/Planting as this has already been done so you can use it for reference, however needs a lot of work since "placement" can be achieved correctly now. (You don't need to follow that same programming style, just use it as a reference for getting fertilizer working and all that)

jasonmoonen:
@Z1Meme I accidentally created this PR with the wrong account, so you can ignore @655456. My discord is: JasonMoonen#0001.
I'll send you a friend request so we can discuss this further.

Issue #790, opened 2022-04-05
groupPackets


QuentinGruber:

QuentinGruber:
maybe useless since there are already grouped packets ( at the app level ) on h1z1 like for weapons

Z1Meme:
@QuentinGruber Is this issue dead?

QuentinGruber:
Not really @Z1Meme i can still do it, i just don't know how to add that to our codebase in a cool and easy way to use. Btw z1br doesn't seems to use this type of internal packet so idk if it will perform well or not (even tho it should)

QuentinGruber:
After several tests, it seems that this feature is not really implemented in the game or works badly. If one day this feature is required to solve performance issues then I would look into it again. ( Daybreak does not use it on its live games )

PR #1065, opened 2022-08-27
Feature/soeserverenh

QuentinGruber:
- [x] Use h1emu-core in unit tests


QuentinGruber:
This version of h1emu-core implement #790 but isn't implemented in soeserver for now

PR #1073, opened 2022-09-06
Command overhaul

Z1Meme:


QuentinGruber:
We should put a deprecation message when you try to use /hax kit for example => "Hi this is deprecated use /kit" for some time. Users would be lost. @Z1Meme

Z1Meme:
> We should put a deprecation message when you try to use /hax kit for example => "Hi this is deprecated use /kit" for some time. Users would be lost. @Z1Meme

Done 👍

Z1Meme:
@QuentinGruber This is ready now

src/utils/utils.ts
```diff
@@ -470,3 +470,16 @@ export function calculateDamageDistFallOff(
   //return damage / (distance * range);
   return damage * Math.pow(range, distance / 10);
 }
+
+export function flhash(str: string) {
```
QuentinGruber, line 474: I will implement this hash func in h1emu-core

QuentinGruber (approved):

PR #1082, opened 2022-10-03
Feature/configFile

QuentinGruber:


Z1Meme (changes requested):
The config loader should check if the enviornment variable is already defined before overwriting it, this way command line env vars overwrite the config file. Also, MONGO_URL probably shouldn't be in the config file since that's sensitive info and the config file will have to be included in docker respositories since you can't really pass a config file into the docker instance easily. Ik you could just add the line yourself and it would still work but it shouldn't be encouraged.

Z1Meme:
> The config loader should check if the enviornment variable is already defined before overwriting it, this way command line env vars overwrite the config file. Also, MONGO_URL probably shouldn't be in the config file since that's sensitive info and the config file will have to be included in docker respositories since you can't really pass a config file into the docker instance easily. Ik you could just add the line yourself and it would still work but it shouldn't be encouraged.

Otherwise looks great

PR #1125, opened 2022-10-23
Feature/split looted stack

Z1Meme:


Z1Meme:
todo: move stack splitting to lootContainerItem instead of pickupItem

QuentinGruber (approved):

PR #1127, opened 2022-10-26
Overhaul lootspawn system

Z1Meme:


Z1Meme:
@QuentinGruber When you have some time please fix these merge errors (they're caused by this branch being old and dev having newer stuff), and merge this so it doesn't conflict with this https://github.com/QuentinGruber/h1z1-server/pull/1131/. All I did so far was convert the old loot tables to a new format which will allow us to do weighted randomization when I figure out that logic.

QuentinGruber (approved):

PR #1175, opened 2022-12-07
WIP worldroutineopti

QuentinGruber:
- [ ] newly added StaticEntities must be created for all clients in the radius in parallel to the worldroutine.


QuentinGruber:
Move to #1196 solution

PR #1183, opened 2022-12-24
Construction Features + Cleanup

Z1Meme:


Z1Meme:
For construction entities that already have an interaction string, holding a demo hammer should show the undo placement option and it should give the full item back. Otherwise, show a "Demolish" message and give some of the recipe components back.

Z1Meme:
https://github.com/QuentinGruber/h1z1-server/issues/1207

QuentinGruber (approved):

PR #1249, opened 2023-01-22
Routines optimization & random fixes

avcio9:


src/servers/ZoneServer2016/classes/gridcell.ts
```diff
@@ -1,8 +1,6 @@
-import { BaseEntity } from "../entities/baseentity";
-
 export class GridCell {
   position: Float32Array;
-  objects: Array<BaseEntity> = [];
+  objects: Array<any> = [];
```
QuentinGruber, line 3: the `unknown `type would be more appropriate

QuentinGruber (dismissed):

src/servers/ZoneServer2016/classes/gridcell.ts
avcio9: The classes are not assignable to unknown type

QuentinGruber: Then we could have a custom type containing all the classes

QuentinGruber (approved):

PR #1336, opened 2023-02-02
fixed server crash from broken character load

Z1Meme:


Z1Meme:
@QuentinGruber This should be a good enough fix for the server crash issue, however I can't seem to reproduce the crash so I don't know if it works 100%. Also I started fixing it one way in the first commit, then while looking back over the code I realized that it should've been done a different way so I reverted.

QuentinGruber (approved):

PR #1342, opened 2023-02-04
Bugfix/reloadfixes

Z1Meme:


QuentinGruber (approved):
Haven't tested it ingame but seems good

PR #1355, opened 2023-02-06
feature/PingLock

avcio9:


QuentinGruber:
So we don't just lock there weapons we kick them? I didn't saw it that way but i'm ok with it. @Z1Meme What do you think of it?

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -168,6 +168,7 @@ export class ZoneServer2016 extends EventEmitter {
   _db!: Db;
   _soloMode = false;
   _useFairPlay = true;
+  _maxPing = 200;
```
QuentinGruber, line 171: should be a public property, servers owner would change this

QuentinGruber (changes requested):

src/servers/ZoneServer2016/zoneserver.ts
avcio9: > So we don't just lock there weapons we kick them? I didn't saw it that way but i'm ok with it. @Z1Meme What do you think of it?

I've talked with @Z1Meme, hes fine with it

QuentinGruber (approved):

Z1Meme:
> So we don't just lock there weapons we kick them? I didn't saw it that way but i'm ok with it. @Z1Meme What do you think of it?

Looks good, we'll see if we need to raise that ping threshold but it'll be fine for now. We should also add a movement lock for sudden ping spikes using update location.

PR #1374, opened 2023-02-08
fixes

avcio9:


avcio9:
This greatly reduces lagg and packet loss when spawning alot of npcs for me

PR #1382, opened 2023-02-09
Fix container error

avcio9:


QuentinGruber (dismissed):

src/servers/ZoneServer2016/entities/constructionchildentity.ts
```diff
@@ -469,6 +469,13 @@ export class ConstructionChildEntity extends BaseLightweightCharacter {
         ...parentFoundation.freeplaceEntities,
         ...this.freeplaceEntities,
```
Z1Meme, line 470: should be removed since you do it correctly below

Z1Meme (changes requested):

src/servers/ZoneServer2016/entities/constructionchildentity.ts
avcio9: done

Z1Meme (approved):

PR #1405, opened 2023-02-13
feature/spectatorTeleport

avcio9:


src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -862,6 +862,12 @@ export class zonePacketHandlers {
       packet
     );
   }
+  SpectatorTP(server: ZoneServer2016, client: Client, packet: any) {
+    server.sendData(client, "ClientUpdate.UpdateLocation", {
+      position: [packet.data.x, 355, packet.data.y, 1],
```
Z1Meme, line 867: Should probably use the same y level that the player is currently at

Z1Meme (changes requested):

src/servers/ZoneServer2016/zonepackethandlers.ts
avcio9: It Has to be that high or player can get teleported under map, will get it fixed along height map

Z1Meme: 👍

Z1Meme (approved):

PR #1440, opened 2023-02-24
Chatmanager

Z1Meme:


src/utils/enums.ts
```diff
@@ -28,6 +28,7 @@ export enum NAME_VALIDATION_STATUS {
 export enum DB_COLLECTIONS {
   ADMINS = "admins",
   BANNED = "banned",
+  MUTED = "muted",
```
QuentinGruber, line 31: Do we will use the same collection for ppl muted on vocal chat ? I know it's not ready but if we plan to separate chat and voice mute let's rename the collection and enum CHAT_MUTED

QuentinGruber (dismissed):

src/utils/enums.ts
Z1Meme: we'll use the same collection I think

QuentinGruber (dismissed):

avcio9 (approved):

PR #1462, opened 2023-02-28
Hi more Skins added To enums :)  

Zam6969:
💜💜💜💜💜💜💜


Zam6969:
Couldn't get the military enum to work for some reason any suggestion?

Zam6969:
Can pull this couldn't figure this out :( commented out the code that doesn't work

src/servers/ZoneServer2016/managers/worldobjectmanager.ts
```diff
@@ -61,6 +63,7 @@ function getRandomSkin(itemDefinitionId: number) {
     Items.HAT_BEANIE,
     Items.HAT_CAP,
     Items.HELMET_MOTORCYCLE,
+    Items.KEVLAR_DEFAULT,
   ];
   if (!allowedItems.includes(itemDefinitionId)) return itemDefinitionId;
```
0-x-2-2, line 68: Why is this check even here when there is a switch statement that could just use the default clause to early return which is faster?

0-x-2-2: The switch statement can be optimized by V8 to be `O(1)` or `O(log n)` if possible, currently, it is checking twice which is an extra O(n) operation that could be `O(1)` or `O(log n)`.

0-x-2-2: It is also possible that this causes extra heap memory allocations but I don't know how good V8 is at doing stack allocation.

Zam6969:
Candles!

Z1Meme (dismissed):

Z1Meme:
@Zam6969 is this ready?

Zam6969:
@Z1Meme  Good?

Z1Meme (approved):

PR #1515, opened 2023-03-18
ConfigManager

Z1Meme:


Z1Meme:
@QuentinGruber Idk how to fix the "Test Binaries" test, I added js-yaml and that seems to be the cause

QuentinGruber:
> @QuentinGruber Idk how to fix the "Test Binaries" test, I added js-yaml and that seems to be the cause

Sorry didn't saw that, will be back tomorow and fix this if it think of it. It's just a "false" require to add so when the exe is build it include the file

Z1Meme:
@QuentinGruber this is ready to merge when the test binaries test is fixed

h1z1-server
QuentinGruber, line 1: ?

package.json
```diff
@@ -13,6 +13,7 @@
     "h1z1-server-demo-2016": "scripts/h1z1-server-demo-2016.js"
   },
   "dependencies": {
+    "@types/js-yaml": "^4.0.5",
```
QuentinGruber, line 16:
```suggestion
    "@types/js-yaml": "4.0.5",
```
Use fixed version ( so it need to redo the package-lock )

QuentinGruber (changes requested):

QuentinGruber:
i'm looking at the binaries test

QuentinGruber (approved):

PR #1518, opened 2023-03-21
simpleNpcs

avcio9:


Z1Meme:
The position / rotation offsets must be fixed before we can merge this, otherwise people will constantly have their placements bugged, and cheaters will just place anywhere

Z1Meme:
Also this would re-enable unraidable base designs that used sideways wall placements on deck foundations / tampers

Z1Meme (changes requested):
Need to fix offsets

Z1Meme (dismissed):

QuentinGruber (approved):

Issue #1401, opened 2023-02-12
Vanish Bug


RywaZ:
when i activate the /vanish and i activate the first person i am blocked i can't remove it anymore

Examples : https://streamable.com/whjdmm

Zam6969:
Same issue :(

BryanMacleod:
@Zam6969 try it and let me know

QuentinGruber:
It's still an issue

QuentinGruber:
well it's due to the fact that we enable spectator mode and it seems to lock to 1st person, if we find a way to still display nametag from huge distance ( wallhack ) then we could remove the use of the spec mode and fix this issue

Issue #1191, opened 2022-12-27
Can't get out of spectate


Z1Meme:
Hitting "E" or doing /spectate while spectating should dismount you from the spectator cam.

QuentinGruber:
This need more knowledge of spectatorBase

Z1Meme:
been fixed for a while

Issue #1286, opened 2023-01-25
/spectate update pos


QuentinGruber:

QuentinGruber:
> @QuentinGruber is this fixed?

seems not when it spectate if you go too far from the pos you started spectating you won't see any netobjects ( since your position isn't updated server-side ) also an atv spawn for some reason why using it now 😅

PR #1531, opened 2023-03-28
Temp Vanish/Spectate fix

BryanMacleod:
Can use for temp fix of #1191 #1401


BryanMacleod:
Also temp fix this issues  #1286

PR #1546, opened 2023-04-02
population numbers workaround

Z1Meme:


Z1Meme:
@QuentinGruber This goes with a launcher update, do not merge without launchers updated first.

src/packets/LoginUdp/LoginUdp_11/loginpackets.ts
```diff
@@ -23,7 +23,8 @@ const serverField: any[] = [
   { name: "descriptionId", type: "uint32" },
   { name: "reqFeatureId", type: "uint32" },
   { name: "serverInfo", type: "string" },
-  { name: "populationLevel", type: "uint32" },
+  { name: "populationNumber", type: "uint16", defaultValue: 0 },
```
QuentinGruber, line 26: Should put a comment about this explaining that we do modify this packet and that it's not is correct structure

src/servers/LoginServer/loginserver.ts
```diff
@@ -226,20 +225,14 @@ export class LoginServer extends EventEmitter {
...
-                      .collection(DB_COLLECTIONS.SERVERS)
-                      .findOne({ serverId: serverId });
                     this._db
                       ?.collection(DB_COLLECTIONS.SERVERS)
                       .findOneAndUpdate(
                         { serverId: serverId },
                         {
                           $set: {
                             populationNumber: population,
-                            populationLevel: getPopulationLevel(
```
QuentinGruber, line 239: if we remove it then remove the util func

QuentinGruber (changes requested):

0-x-2-2:
It could be possible to detect if this is an old client or not by changing the protocol version for custom protocol features supported or maybe just using a custom packet entirely.

QuentinGruber (approved):

PR #1585, opened 2023-05-05
Fixed various exploits with door locks.

0-x-2-2:


QuentinGruber (approved):
Nice @0-x-2-2 , would be cool if you add a proper title to this PR :cowboy_hat_face:

0-x-2-2:
The title was not included as it is related to base security. But it fixes various exploits related to cracking/brute-forcing base codes.

PR #1609, opened 2023-07-25
Added links to the Thanks List

thegrreat1:
Still a few one missing that i couldnt find the link to. The reason i added the links is that i think we should do proper credits for the ones that actually helped and not just put the names ;)


QuentinGruber (approved):
Good idea ! 🤠

PR #1612, opened 2023-08-14
feature/vehicleLocks

Nikolassparrow:
I messed up the first pull request so I made another to clean it up lol. Anyways:

I've been wanting to contribute to this project since I've joined, but took some time to learn the logic and gain the confidence to implement a meaningful change. This PR is for vehicle locks which solves #526.


src/packets/ClientProtocol/ClientProtocol_1080/vehicle.ts
```diff
@@ -195,7 +195,20 @@ export const vehiclePackets: PacketStructures = [
...
+  "Vehicle.AccessType", 
+  0x891c,
+  {
+    fields: [
+      {
+        name: "vehicleGuid",
+        type: "uint64string",
+        defaultValue: "0x000000000000000000"
+      },
+      { name: "accessType", type: "uint16", defaultValue: 0 }
```
QuentinGruber, line 208: What are the different accessTypes ?Llike 0 is open 2 is locked ? What is 1 and the other ones? Beside that it looks good will just test it in game and it will be merge, nice one ! :cowboy_hat_face:

QuentinGruber:
I will look at the TS issue when i will test it ingame ;)

QuentinGruber (dismissed):

src/packets/ClientProtocol/ClientProtocol_1080/vehicle.ts
Nikolassparrow: I should've been more specific, but yes from what I've tested it seems 2 is locked and every other number : 1, 3, 4, 5, 6, 7, 8, 9 are all unlocked. I couldn't find another use for the other numbers but it seems like 0 and 2 are the "main" numbers.

QuentinGruber: Maybe one of the number also check the "team" the vehicle is on or smth like that ( since it's ps2 stuff )

PR #1640, opened 2023-09-21
bugFixes

avcio9:


Z1Meme:
@avcio9 Storages should be prevented from placement above a tree stump instead of destroyed when a tree respawns in case someone places a storage there on accident. Would prevent people from losing their stuff from a simple mistake. Otherwise everything looks good.

PR #1644, opened 2023-09-24
Weather Rework

Z1Meme:


Z1Meme:
merging and closing this for now until we decide to do a full on weather rework

PR #1646, opened 2023-09-25
buh

Zam6969:
adding more  tho


Zam6969:
how can i target self if only /kit pvp is typed

Zam6969:
Got it nvm

Zam6969:
done pull branch whenever

Zam6969:
basically it will get where ur corsair is at and teleport u there

Zam6969:
but help

QuentinGruber (dismissed):
/tpl always tp me under the map and /kit doesn't work, probably bc there is a duplicate

Zam6969:
@QuentinGruber i need help with the tpl command and the kit command should work now

Zam6969:
doesnt work correctly cant figure it out
i tried this 

  {
    name: "tttt",
    permissionLevel: PermissionLevels.ADMIN,
    execute: (server, client, packetData: any) => {
      server.sendData(client, "ClientUpdate.UpdateLocation", {
        position: packetData.position,
      });
    },
  },
  
  doesnt work also

Zam6969:
@QuentinGruber  can you help me with the tpl command i cant figure it out.

Zam6969:
so gkit works but doesnt give it to urself if u type gkit pvp you would have to type gkit pvp zammu

Z1Meme:
Removed /gkit since there's really no reason an admin would give a normal player the pvp/build/parts/skins kit unless on a modded server, in which case they could add that command themselves. Also removed /tpl since it can't work without a serversided heightmap, which is not added yet. Would be better to do this with a special admin gun that TPs the player to the hitLocation of the projectile.

Z1Meme (approved):

Z1Meme (approved):

PR #1647, opened 2023-09-26
hud indicators

avcio9:


Z1Meme (changes requested):
- Bleeding doesn't work (Idk if it's broken in general or what but I can't trigger it to test the bleeding icon)
- Default value for tiredness should be max so you don't have to sleep as soon as you spawn
- Sleeping mats don't work so there's no real way to sleep that I can find
- Coffee / Coffee with sugar doesn't seem to affect tiredness (The message still shows up after drinking multiple coffees)
- Fall damage is broken ? Maybe it's an issue on dev from the ClientSettings name change

Z1Meme:
Swizzle icon is really the only one I could get to show, idk what the other icons are.

Z1Meme (approved):

PR #1652, opened 2023-10-10
Redo ban system / Asset validation

Z1Meme:


Z1Meme:
@QuentinGruber I fully tested this but I need a second opinion when you have time. Old zoneservers will still allow connections with the updated loginserver, but new bans won't work, only ones already registered. The main thing is that the loginserver no longer handles any bans beside global, it's handled on the zone's side entirely, and the loginserver gets that information from the zone before allowing a connection.

QuentinGruber:
> @QuentinGruber I fully tested this but I need a second opinion when you have time. Old zoneservers will still allow connections with the updated loginserver, but new bans won't work, only ones already registered. The main thing is that the loginserver no longer handles any bans beside global, it's handled on the zone's side entirely, and the loginserver gets that information from the zone before allowing a connection.

checking that tomorow :)

QuentinGruber:
We will probably need to modify stuff in live-login-server https://github.com/H1emu/live-login-server

src/packets/LoginUdp/LoginUdp_11/loginpackets.ts
```diff
@@ -377,6 +377,28 @@ const packets: PacketStructures = [
         { name: "message", type: "string", defaultValue: "" }
       ]
     }
+  ],
+  [
+    "H1emu.HadesInit",
```
QuentinGruber, line 382: Hades ?

QuentinGruber:
> @QuentinGruber I fully tested this but I need a second opinion when you have time. Old zoneservers will still allow connections with the updated loginserver, but new bans won't work, only ones already registered. The main thing is that the loginserver no longer handles any bans beside global, it's handled on the zone's side entirely, and the loginserver gets that information from the zone before allowing a connection.

For me it's good and i don't care that this update isn't backward compatible with old zoneservers. The fact that the zoneserver is the one who decide is probably a better idea, but how this will work for ip bans ? Or hwid bans that were on the login server side ? We bundle that all in "global bans" ?

Z1Meme:
> > @QuentinGruber I fully tested this but I need a second opinion when you have time. Old zoneservers will still allow connections with the updated loginserver, but new bans won't work, only ones already registered. The main thing is that the loginserver no longer handles any bans beside global, it's handled on the zone's side entirely, and the loginserver gets that information from the zone before allowing a connection.
> 
> For me it's good and i don't care that this update isn't backward compatible with old zoneservers. The fact that the zoneserver is the one who decide is probably a better idea, but how this will work for ip bans ? Or hwid bans that were on the login server side ? We bundle that all in "global bans" ?

HWID and IP bans are already handled separately from global, the loginserver tells the zone if a client is ip or hwid banned or using a vpn, and the zone tells the login if it wants to accept that client or not.

src/packets/LoginUdp/LoginUdp_11/loginpackets.ts
Z1Meme: That's what Jamie named the asset validation stuff, HadesInit isn't really used rn so we just use HadesQuery to start the validation as soon as a client connects to the login.

Z1Meme:
> We will probably need to modify stuff in live-login-server https://github.com/H1emu/live-login-server

This should be fine, isClientVerified / HwidBanned didn't change and still serve the same purpose.

avcio9:
@QuentinGruber Looking for contact, can you check discord?

QuentinGruber:
@avcio9 yep i will be back on discord in like 2-3h ( i don't have it on my phone )

PR #1666, opened 2023-11-03
Ammo box loot action

jasonmoonen:
^^ title


QuentinGruber:
Hey @jasonmoonen thx for your contribution, we have a release tonight after that we could review your PR and merge it :cowboy_hat_face:

src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -2256,6 +2256,10 @@ export class ZonePacketHandlers {
           "Parts may be required. Open vehicle loadout."
         );
         break;
+      case ItemUseOptions.UNPACK:
+        server.useAmmoBox(client, item);
```
QuentinGruber, line 2260: If there were one in the original game then yes

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -6106,6 +6106,22 @@ export class ZoneServer2016 extends EventEmitter {
     }, 600000);
   }
 
+  useAmmoBox(client: Client, item: BaseItem) {
+    if (!this.removeInventoryItem(client.character, item)) return;
+
+    // Configure amount of ammo? Should be 60 by default.
```
QuentinGruber, line 6112: Yep should be in the config file

Z1Meme: Ammo boxes weren't used in this version and were leftover from 2015 / maybe early 2016, so if these are used it'll proabably only be on modded servers so idk if we should put it in config or not. Pistol bullets have the same model as the ammo box so it'd be confusing to spawn them with eachother.

src/servers/ZoneServer2016/zonepackethandlers.ts
Z1Meme: There was

Z1Meme (approved):

Z1Meme:
@jasonmoonen Merging now, thanks!

PR #1686, opened 2023-11-06
PVE

QuentinGruber:
Known remaining pvp actions:
* We can still pop helmets and armor
* Traps are still active

Need also to make isPve configurable in the config file


Z1Meme:
@QuentinGruber I redid a lot of explosives logic in #1691, and it'll probably conflict with this branch until it's merged

QuentinGruber:
@Z1Meme it's hard to review but for me i didn't change anything that you made in your previous PR but if you wanna take a look just to be sure. ( we can revert it if it's the case )

PR #1690, opened 2023-11-07
Cache static DTO list on server start

Z1Meme:


src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -3168,50 +3172,56 @@ export class ZoneServer2016 extends EventEmitter {
     }
   }
 
-  customizeDTO(client: Client) {
-    const DTOArray: Array<any> = [];
+  customizeStaticDTOs() {
+    console.time("customizeStaticDTOs");
```
QuentinGruber, line 3176:
```suggestion
```

QuentinGruber (approved):

src/servers/ZoneServer2016/zoneserver.ts
Z1Meme: This only happens on server start so I think it's fine to have the time there, the more objects we add (ex. fences) will add to this time so it's good to see how long it takes

PR #1704, opened 2023-11-12
Worn letters!

jasonmoonen:
There's a 10% chance to receive a worn letter while killing zombies by default. Rewards are currently static, except for a 50% chance of getting a shotgun or an AR-15.


src/servers/ZoneServer2016/entities/treasurechest.ts
```diff
@@ -0,0 +1,132 @@
...
+                guid: this.characterId,
+                stringId: StringIds.OPEN
+            });
+        }
+    }
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    OnPlayerSelect(
+        server: ZoneServer2016,
+        client: ZoneClient2016,
+        isInstant?: boolean
```
QuentinGruber, line 112: Don't let unused vars :cowboy_hat_face:

src/servers/ZoneServer2016/managers/worldobjectmanager.ts
```diff
@@ -342,6 +345,28 @@ export class WorldObjectManager {
   }
 
   createLootbag(server: ZoneServer2016, entity: BaseFullCharacter) {
+    if (entity instanceof Zombie) {
+      //TODO: Probably should rework this?
+      const wornLetters = [
```
QuentinGruber, line 350: wornLetters could be a static var of worlObjectManager

QuentinGruber (commented):
Thx that's awesome !

QuentinGruber:
I've tested it ingame and it doesn't work for me

jasonmoonen:
> I've tested it ingame and it doesn't work for me

Did you update the config file? I’ve had this happen while adding the config for water sources.

if the config is not updated the variable will be undefined by default, so it won’t give rewards.

Maybe it’s a good idea to implement a config versioning and upgrade the older config acordingly?

QuentinGruber:
> > I've tested it ingame and it doesn't work for me
> 
> Did you update the config file? I’ve had this happen while adding the config for water sources.
> 
> if the config is not updated the variable will be undefined by default, so it won’t give rewards.
> 
> Maybe it’s a good idea to implement a config versioning and upgrade the older config acordingly?

Ok now i see them but can't interact with the lootbag, and config versioning is kinda already a thing.

jasonmoonen:
> > > I've tested it ingame and it doesn't work for me
> > 
> > 
> > Did you update the config file? I’ve had this happen while adding the config for water sources.
> > if the config is not updated the variable will be undefined by default, so it won’t give rewards.
> > Maybe it’s a good idea to implement a config versioning and upgrade the older config acordingly?
> 
> Ok now i see them but can't interact with the lootbag, and config versioning is kinda already a thing.

Hey @QuentinGruber You should be able to take the worn letters now

QuentinGruber:
> > > > I've tested it ingame and it doesn't work for me
> > > 
> > > 
> > > Did you update the config file? I’ve had this happen while adding the config for water sources.
> > > if the config is not updated the variable will be undefined by default, so it won’t give rewards.
> > > Maybe it’s a good idea to implement a config versioning and upgrade the older config acordingly?
> > 
> > 
> > Ok now i see them but can't interact with the lootbag, and config versioning is kinda already a thing.
> 
> Hey @QuentinGruber You should be able to take the worn letters now

yep it works now !

PR #1715, opened 2023-12-01
Emote 35 exploit fix

Nikolassparrow:
https://streamable.com/d76l15

someone in NTC told me that 1k had found this, but never reported it (more than likely abusing the hell out of it). Simple fix that checks animationId for 35 and if so exits the function.


Z1Meme:
Thanks @Nikolassparrow

QuentinGruber:
@Z1Meme I didn't saw that, but verify that the PR target /dev when you're reviewing

PR #1717, opened 2023-12-02
plugin fixes

Z1Meme:


Z1Meme:
@QuentinGruber It's ready enough for now, will do some more plugin work later.

PR #1738, opened 2023-12-17
Added loaded modules command

jamieduke:
Small little AC measure, very similar to listprocesses, but for loaded modules. This is a little bit time-critical due to recent events.

This time it's on the right branch.


QuentinGruber:
Hey @jamieduke thx ! just run `npm run gen-packets-interfaces` to fix the last warning, i can't do it since it's on your fork.

PR #1752, opened 2023-12-25
explosiveCandleGlow

Nikolassparrow:
Currently when you ignite an IED, the first explosive will passthrough igniteoptionpass and exit on the first go-around, so I took out the return and made the radius the same as furnaces. this worked on singleplayer but should probably be tested before being pushed to live.


QuentinGruber (approved):
For me it seems OK, and singleplayer vs multiplayer should not change anything on that feature so as you want if you wanna test more :)

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -6790,15 +6790,16 @@ export class ZoneServer2016 extends EventEmitter {
...
+          2.0,
           client.character.state.position,
-          this._explosives[a].state.position
+          explosive.state.position
         )
       ) {
-        this._explosives[a].ignite(this, client);
-        return;
+        explosive.ignite(this, client);
+        Scheduler.wait(750);
```
QuentinGruber, line 6802:
```suggestion
        Scheduler.wait(750);
```
why do you wanna await that much ?

Nikolassparrow: I wanted to make it so the IED's light 1 by 1, testing it some more though and it didn't even work, just delayed the ignition process. Gonna look into it some more

QuentinGruber (approved):

PR #1758, opened 2023-12-27
Dressed Bandage recipe

Nikolassparrow:
Reported by "The Rizzler" on discord. Dressed bandage recipe didn't give back the bottle when crafted, in his words: "you're emptying the honey bottle onto the dressing so it's weird that the bottle would be consumed in the process." Fix: return an empty bottle lol


QuentinGruber (approved):

QuentinGruber:
nice !

Issue #1753, opened 2023-12-25
Hand Shovel = No Permissions ?!


DaEgo:
Hand Shovel create a Model but its useless
Cant open it
Cant destroy it with Demo Hammer

![Hand Shovel](https://github.com/QuentinGruber/h1z1-server/assets/1351055/7b5bdc34-0374-457b-ab57-d03c5bf74681)

QuentinGruber:
Thx for the report @DaEgo , will look at that when i'm done on #1721 :cowboy_hat_face:  ( but if others want to, they can )

PR #1759, opened 2023-12-28
Add hand shovels

Nikolassparrow:
Fixed and added hand shovels back. The storage is unlimited right now but could probably be fine tuned to allow only for a small amount of items (like how it was back in 2016). Fix for #1753.


src/servers/ZoneServer2016/managers/constructionmanager.ts
```diff
@@ -1640,6 +1650,50 @@ export class ConstructionManager {
...
+      modelId,
+      position,
+      rotation,
+      server,
+      scale,
+      itemDefinitionId,
+      parentObjectCharacterId || "",
+      ""
+    );
+    console.log(obj);
```
QuentinGruber, line 1676:
```suggestion
```

QuentinGruber (approved):

PR #1762, opened 2023-12-30
bugfix/npcReplication

### LINKED ISSUE #1673: Item on player should not glow
image

Nikolassparrow:
Essentially the fix for #1750 and #1733 only worked because the function never even got down to Replication.NpcComponent. Objects still need to be sent the packet so instead revert the change and move the logic down the function. This fixes #1673 for good.


QuentinGruber (approved):
Gg

PR #1765, opened 2024-01-03
Add barbed wire

Nikolassparrow:
Added barbed wire into the game, the barbed wire cube bounds was a bit wonky to work with so could need adjustment, but I'm confident that this cube now is a good one. Also noticed while working on this that TrapEntity props that spawn around the map (like PV church), once destroyed will never work again. There should be a detection on whether a SpikeTrap or BarbedWire is a prop or not so they can't be destroyed.


src/servers/ZoneServer2016/managers/constructionmanager.ts
```diff
@@ -1346,6 +1347,7 @@ export class ConstructionManager {
       );
     npc.arm(server);
     server._traps[characterId] = npc;
+    console.log(server._traps[characterId]);
```
QuentinGruber, line 1350:
```suggestion
```

QuentinGruber (approved):

PR #1766, opened 2024-01-03
Bugfix with melee bleeding on PvE

Nikolassparrow:
Was reported by a user name Triggl/ Bornat on discord. The check for PvE doesn't occur until the damage function, so add a check before the bleeding code can occur to prevent players from killing each other due to bleeding.


src/servers/ZoneServer2016/entities/character.ts
jasonmoonen, line 1: This will make zombies not harvestable. Also add a check if they're a NPC or not because they should be able to receive damage

QuentinGruber (approved):

PR #1767, opened 2024-01-05
Require skinning knife for harvesting

Nikolassparrow:
Follow-up for the lootChanges branch, users should spawn with a Jagged knife and should be checked when harvesting NPC's. Also, DBG removed melee hitting for harvesting back in 2015, they instead used an OnInteractionString to harvest animals in 2016 so that will be the next move for harvesting.


QuentinGruber:
this will be in 0.30.1 :)

QuentinGruber:
Don't seems to work for me, could you provide a video about that ?

Nikolassparrow:
> Don't seems to work for me, could you provide a video about that ?

https://streamable.com/j3iz24

I also noticed while testing this that depending on what weapon you use to kill the NPC, it will allow you to get in more "harvest hits", for example: I kill a zombie with a shotgun in the head and I can only hit them once with a harvest tool, where-as I use a knife to kill a zombie and get 4-5 "harvest hits" compared to the 1 with the shotgun kill. There's a bug with how NPC's receive damage and their subsequent harvest health afterwards

QuentinGruber:
This is still relevant ? @Nikolassparrow

Nikolassparrow:
> This is still relevant ? @Nikolassparrow

Nope! You merged the animal harvesting branch so this one can be closed now 👍

PR #1771, opened 2024-01-05
Barbed wire damaging players in vehicle

Nikolassparrow:
Reported by @DaEgo barb wire should dmg players inside of vehicles


jasonmoonen:
Shouldn’t the barbed wire damage the vehicle instead?

Nikolassparrow:
> Shouldn’t the barbed wire damage the vehicle instead?

Was this how it was back in the day? I'll do some more research and if so add it in

DaEgo:
how i remember, it was a solid flat "wall" block
u can jump on/over it but does damage
and cars cant (of course) not pass

QuentinGruber (approved):

PR #1772, opened 2024-01-05
Fix for wrench not working on PVE

### LINKED ISSUE #1744: Wrench does nothing on Vehicles
You can hit a vehicle with the wrench,
it also loses durability.
Nothing happens with the Vehicle, just the Wrench broke at the End.

Nikolassparrow:
^^
This fixes #1744

### DIFF
--- a/src/servers/ZoneServer2016/entities/vehicle.ts
+++ b/src/servers/ZoneServer2016/entities/vehicle.ts
@@ -484,7 +484,6 @@ export class Vehicle2016 extends BaseLootableEntity {
   }
 
   damage(server: ZoneServer2016, damageInfo: DamageInfo) {
-    if (server.isPvE && damageInfo.hitReport?.characterId) return;
     if (this.isInvulnerable) return;
     const oldHealth = this._resources[ResourceIds.CONDITION];
     this._resources[ResourceIds.CONDITION] -= damageInfo.damage;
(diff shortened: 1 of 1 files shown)


jasonmoonen:
The wrench shouldn’t be fixed on PVE servers. You need to prevent it from spawning if the server is PVE.
People would farm cars like they did in the original game, hence Daybreak not letting wrenches spawn on PVE servers at some point.

Nikolassparrow:
> The wrench shouldn’t be fixed on PVE servers. You need to prevent it from spawning if the server is PVE. People would farm cars like they did in the original game, hence Daybreak not letting wrenches spawn on PVE servers at some point.

https://i.gyazo.com/5daf2022465e75a5d64da66acaed8f6f.png
https://i.gyazo.com/e98825e85f78707770f6bd311f62af78.png

Wrenches were removed from PvE in the August 2017 update as well as the February 2018 update. December 2016 had wrenches spawn on PvE but wrench spawn rates need be decreased for pvp and pve dramatically. From what I've researched, DBG in 2016/2015 consistently lowered spawn rates on wrenches to prevent vehicle hoarding so that should be looked into further, not removing the entire ability to repair cars (this PR also fixes players damaging vehicles too).

QuentinGruber (approved):

PR #1773, opened 2024-01-06
Hand shovel changes

Nikolassparrow:
Got this info from KronicWayz 👍 hand shovel stashes shouldn't glow and also change the name_id to 1698 which makes it say "open stash" rather than "open hand shovel"

### DIFF
--- a/src/servers/ZoneServer2016/zonepackethandlers.ts
+++ b/src/servers/ZoneServer2016/zonepackethandlers.ts
@@ -1776,7 +1776,8 @@ export class ZonePacketHandlers {
       if (
         !(
           entity instanceof ConstructionParentEntity ||
-          entity instanceof ConstructionChildEntity
+          entity instanceof ConstructionChildEntity ||
+          entity instanceof LootableConstructionEntity
         )
       ) {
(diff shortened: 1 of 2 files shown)


QuentinGruber (approved):

src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -1776,7 +1776,8 @@ export class ZonePacketHandlers {
       if (
         !(
           entity instanceof ConstructionParentEntity ||
-          entity instanceof ConstructionChildEntity
+          entity instanceof ConstructionChildEntity ||
+          entity instanceof LootableConstructionEntity
```
QuentinGruber, line 1780: this won't affect other stuff ?

Nikolassparrow: Not from what I've tested, looks good so far 👍

jasonmoonen: Did you test decay? Because objects not placed on a foundation will decay a lot faster

Issue #1693, opened 2023-11-07
Shorten and brighten nights


Z1Meme:
- Need to shorten the length of nights since night seems to be longer than day right now
- Brighten night since the lighting update made it much darker (Should still be darker than before v0.29 though)
- Server start should always be day

PR #1779, opened 2024-01-08
WIP broken stuff on servertime vs gametime vs ingametime

QuentinGruber:
I want to investigate this more since something feels weird about how we handle time in general on the server


QuentinGruber:
https://github.com/QuentinGruber/h1z1-server/issues/1693

PR #1795, opened 2024-01-12
Grief foundations

Nikolassparrow:
d60a86853b5c95822647a369e174dd3a
69ed938204b852d82b91d6f272e8485a
7120197a35ad9b025b0b041c348d3339

Currently bases that only have a few gates clutter the map and a quick 5 minute walk on any server will let you see this clearly. I added a grief check every DecayTick to check if foundations (tampers or deck) have less than 4 wall slots (so AT LEAST 4 gates to avoid anti-grief), a minimum of 1 shelter, or expansions to prevent starter bases getting this debuff (although it will go away if you are actively maintaining and building up your base). I set the decay timer to decay grief bases after 3 days as currently wipes are only 2 weeks so this should make this change meaningful, but in April 2016, DBG made it so grief bases decayed in 7 days. Todo: eventually allow people to claim grief bases rather than destroying them (when claiming bases becomes available).


QuentinGruber (approved):

QuentinGruber:
Was thinking about it (@Nikolassparrow) this would work only if the server keep running for 3 days but they currently auto restart every two days on live 😅

QuentinGruber:
so we may have to store that information or using the creation timestamp from mongo

QuentinGruber:
https://www.mongodb.com/docs/manual/reference/method/ObjectId.getTimestamp/

QuentinGruber (changes requested):
comments

Nikolassparrow:
> Was thinking about it (@Nikolassparrow) this would work only if the server keep running for 3 days but they currently auto restart every two days on live 😅

I changed up the logic to where it should work despite restarts, I'm not too familiar with mongo, but I could look into it but as of now this SHOULD work.

src/servers/ZoneServer2016/managers/decaymanager.ts
```diff
@@ -61,6 +64,37 @@ export class DecayManager {
...
+        foundation.itemDefinitionId == Items.GROUND_TAMPER
+      ) {
+        if (
+          Object.keys(foundation.occupiedWallSlots).length <=
+            this.griefCheckSlotAmount &&
+          Object.keys(foundation.occupiedShelterSlots).length == 0 &&
+          Object.keys(foundation.occupiedExpansionSlots).length == 0
+        ) {
+          if (
+            Date.now() - foundation.creationTime >=
```
QuentinGruber, line 81: This if statement could be done before the one above so we could save cpu time since this statement is a small comparison instead of creating an array of keys to check length ( array that would be deleted after this stack also )

src/servers/ZoneServer2016/entities/constructionparententity.ts
```diff
@@ -72,6 +72,7 @@ export class ConstructionParentEntity extends ConstructionChildEntity {
   readonly rampSlots: ConstructionSlotPositionMap = {};
   occupiedRampSlots: { [slot: number]: ConstructionChildEntity } = {};
   lastDamagedTimestamp: number = 0;
+  creationTime = Date.now();
```
QuentinGruber, line 75:
```suggestion
  readonly creationTime = Date.now();
```

data/2016/sampleData/defaultconfig.yaml
```diff
@@ -129,6 +129,8 @@ decay:
 
   vehicleDamageTicks: 3 # damage vehicles every hour
   vacantFoundationTicks: 3 # destroy empty foundations after 1 hour
+  griefFoundationTimer: 259200000 # destroy grief foundations after 3 days
```
QuentinGruber, line 132: could be in hours instead of ms, users don't need that level of precision

QuentinGruber (commented):
small suggestions, I'll test it later today :)

src/servers/ZoneServer2016/managers/decaymanager.ts
QuentinGruber, line 81: also placementTime already exist and would be cool to use it so we don't need to wipe or do hacks for this to work

src/servers/ZoneServer2016/managers/decaymanager.ts
```diff
@@ -61,6 +64,37 @@ export class DecayManager {
     }, this.decayTickInterval);
   }
 
+  private constructionGriefCheck(server: ZoneServer2016) {
```
QuentinGruber, line 67: Btw why not doing that stuff at the beginning of contructionExpirationCheck method ? to avoid looping two time in a row on the same data?

data/2016/sampleData/defaultconfig.yaml
Nikolassparrow: #1859

src/servers/ZoneServer2016/managers/decaymanager.ts
Nikolassparrow: #1859

QuentinGruber:
dup

PR #1821, opened 2024-01-18
Add Animal harvesting without melee and Add food poisoning

Nikolassparrow:
https://www.youtube.com/watch?v=1QxoV3D-uSI&t=116s
https://www.youtube.com/watch?v=-l2DSMGLmoo (funny video of harvesting)
https://www.youtube.com/watch?v=PJsc38r_9uw (food poisoning video)

In the July 2016 update, DBG removed melee harvesting and added harvesting by interaction. This PR adds the following: 

- Interaction harvesting on npcs, removes old melee system
- Removed deer scents from harvest and slightly increase animal fat rates
- Change logic to where at least 1 item is generated from each harvest (because why wouldn't a fully grown bear not give at least one item 😂)
- Add food poisoning hudindicator and more foods to be poisonous. (in december 2016, poisonous foods only did 3 damage and it was in increments of 25 damage each, so 12 in total).

Known issues:
There should be an animation to go along with harvesting (shown in the 2nd video above ^) but I wasn't able to find an animation list or anything from doing /emote.


src/servers/ZoneServer2016/entities/character.ts
```diff
@@ -957,7 +971,7 @@ export class Character2016 extends BaseFullCharacter {
       oldHealth = this._resources[ResourceIds.HEALTH];
     if (!client) return;
 
-    if (this.isGodMode() || !this.isAlive || damage < 100) return;
+    if (this.isGodMode() || !this.isAlive || damage < 25) return;
```
QuentinGruber, line 974:
```suggestion
    if (this.isGodMode() || !this.isAlive || damage <= 25) return;
```
?

src/servers/ZoneServer2016/entities/npc.ts
```diff
@@ -17,16 +17,17 @@ import { BaseFullCharacter } from "./basefullcharacter";
...
-import { Items, MeleeTypes, StringIds } from "../models/enums";
+import { Items, StringIds } from "../models/enums";
 import { CommandInteractionString } from "types/zone2016packets";
 
 export class Npc extends BaseFullCharacter {
   health: number;
   npcRenderDistance = 80;
   spawnerId: number;
   deathTime: number = 0;
+  isZombie = false;
```
QuentinGruber, line 28: Not a huge fan of this, zombies / wildlife should have their own class extending npc one or at least an Enum that replace the 2 booleans

src/servers/ZoneServer2016/entities/npc.ts
```diff
@@ -89,15 +90,19 @@ export class Npc extends BaseFullCharacter {
...
+
+          if (!server._soloMode) {
+            logClientActionToMongo(
+              server._db.collection(DB_COLLECTIONS.KILLS),
+              client,
+              server._worldId,
+              { type: killType }
+            );
+          }
+          (client.character.metrics as any)[killType + "Killed"]++;
```
QuentinGruber, line 104: Nasty :P , well i would prefer an if or a switch it will almost add no processing and will have type safety

src/servers/ZoneServer2016/entities/npc.ts
```diff
@@ -232,73 +202,136 @@ export class Npc extends BaseFullCharacter {
...
+        this.isWildlife = true;
         break;
     }
   }
 
+  OnPlayerSelect(server: ZoneServer2016, client: ZoneClient2016) {
+    if (!this.isAlive && client.character.hasItem(Items.SKINNING_KNIFE)) {
+      server.utilizeHudTimer(client, this.nameId, 5000, 0, () => {
+        switch (this.actorModelId) {
+          case 9510:
```
QuentinGruber, line 261: We don't have an enum for them ? If not we should do one

src/servers/ZoneServer2016/entities/zombie.ts
```diff
@@ -37,34 +36,4 @@ export class Zombie extends Npc {
     );
     this.materialType = MaterialTypes.ZOMBIE;
   }
-
-  OnInteractionString(server: ZoneServer2016, client: ZoneClient2016) {
```
QuentinGruber, line 41: it wasn't used ?

src/servers/ZoneServer2016/entities/character.ts
```diff
@@ -449,6 +450,19 @@ export class Character2016 extends BaseFullCharacter {
         server.sendHudIndicators(client);
       }
     }
+    const indexFoodPoison =
+      this.resourceHudIndicators.indexOf("FOOD POISONING");
```
QuentinGruber, line 454: use .includes()

QuentinGruber (changes requested):

src/servers/ZoneServer2016/entities/character.ts
Nikolassparrow: Messed with this and wasn't able to get the index of the current position of the food poisoning without having to put in some wonky code for it. Also all other resourceHudIndicator calls are using indexOf too.

src/servers/ZoneServer2016/entities/zombie.ts
Nikolassparrow: I tried to use the Zombie OnInteractionString, but it wouldn't work with my Npc OnInteractionString since zombie extends from Npc, so I just combined them all temporarily in the npc class, once each individual animal class is setup, this can be revised to each class having its own OnInteractionString and OnPlayerSelect.

src/servers/ZoneServer2016/entities/npc.ts
Nikolassparrow: Done 👍

src/servers/ZoneServer2016/entities/npc.ts
Nikolassparrow: Yea, when writing that I didn't like it but it was to reduce redundant code really, I changed it to something more type safe :P

src/servers/ZoneServer2016/entities/character.ts
Nikolassparrow: You can't actually damage someone for 100 damage so that's why I lowered it to 25 so I could send 25 damage to the player. Could also just make it so the player takes 100 damage over 4 intervals so this change doesn't have to be made, but it was mainly made to be accurate with the old H1Z1 food poisoning.

src/servers/ZoneServer2016/entities/npc.ts
Nikolassparrow: I edited this to use the NpcIds enums instead.

src/servers/ZoneServer2016/entities/character.ts
Nikolassparrow: Idk how but the code still worked 😂

src/servers/ZoneServer2016/entities/npc.ts
```diff
@@ -75,7 +75,7 @@ export class Npc extends BaseFullCharacter {
   damage(server: ZoneServer2016, damageInfo: DamageInfo) {
     const client = server.getClientByCharId(damageInfo.entity),
       oldHealth = this.health;
-
+    console.log(client);
```
QuentinGruber, line 78:
```suggestion
```

jasonmoonen (approved):

QuentinGruber (approved):

QuentinGruber:
### Merge activity

* **Jan 31, 9:50 AM**: @QuentinGruber merged this pull request with [Graphite](https://app.graphite.dev/github/pr/QuentinGruber/h1z1-server/1821).

PR #1822, opened 2024-01-19
JS sku

avcio9:
Fixes every sound in the game (menu, poi, zombies)
Fixes blackberry pickups,
- Possibly more

Edited raw Z1BR login reply packet and sending it instead of packing dataschema. 
It seems like the struct that we have is incorrect, but for now im not into reversing it :)


QuentinGruber:
It also fixed missing stuff in the marketplace

QuentinGruber:
also displaying airdrop tickets

QuentinGruber:
So maybe my work on account item works now

QuentinGruber (approved):

PR #1844, opened 2024-01-24
Reduce spawn timer on low / medium pop

QuentinGruber:
Maybe this is too much, but rn ppl don't play on low pop since the loot is so bad. I know we done that to avoid unemployed to have too much of an advantage but the fact is there is a lot of them. At the beginning of every wipe the server pop is always 40+ even at night, then these ppl have their bases and just stop playing and others who want to start the wipe mid train just can't due to low loot and that make a snowball and nobody play anymore.

### DIFF
--- a/src/servers/ZoneServer2016/managers/worldobjectmanager.ts
+++ b/src/servers/ZoneServer2016/managers/worldobjectmanager.ts
@@ -195,21 +195,12 @@ export class WorldObjectManager {
 
     const playerCount = _.size(server._characters);
 
-    switch (true) {
-      case playerCount <= 20:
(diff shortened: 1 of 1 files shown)


QuentinGruber:
Current dependencies on/for this PR:
* `dev`
  * **PR #1844** <a href="https://app.graphite.dev/github/pr/QuentinGruber/h1z1-server/1844?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/QuentinGruber/h1z1-server/1844?utm_source=stack-comment).

PR #1859, opened 2024-01-25
Faster decay on grief plots

Nikolassparrow:
#1795 was on a forked branch, made one on this repository to be able to edit with but couldn't figure out how to merge this pull request with the original lol. Made some new changes where hrs are used in the config as-well as update some of the logic so we aren't looping through the same data twice.


Nikolassparrow:
Any update on this? Should be all good to push to dev 👍

src/servers/ZoneServer2016/managers/decaymanager.ts
```diff
@@ -64,6 +66,32 @@ export class DecayManager {
...
+            Object.keys(foundation.occupiedExpansionSlots).length == 0
+          ) {
+            for (const a in foundation.occupiedWallSlots) {
+              foundation.occupiedWallSlots[a].destroy(server);
+            }
+            // clear floating entities
+            for (const a in foundation.freeplaceEntities) {
+              foundation.freeplaceEntities[a].destroy(server);
+            }
+            foundation.destroy(server);
```
QuentinGruber, line 90: Maybe just add a log if a foundation is destroyed due to this system so we can see how often it's trigerred

Nikolassparrow: Currently it's 3 days, but i'll add the log for it 👍🏿

QuentinGruber: Yep but more to see how many grief bases were deleted 🙂

PR #1865, opened 2024-01-26
bugfix/furnaceBurnTime

Nikolassparrow:
There was an issue with furnaces on live servers and the only changes recently were from the burnTime function that I implemented. Something gets messed up when using that function, so I decided to try and branch off from the if statement that was already there. So far there hasn't been issues while testing it on singleplayer, but the burnTime change could have to be reverted until more testing can be done if this change doesn't fix it. 

Update: So I had this bug happen to my furnaces and they didn't work for a while, but randomly after a few hours they started too again...? I don't know if this issue is related to burnTime... so this pr could be pointless, but I'll let the current furnaces on main be tested a bit more before I can 100% say this PR will be the fix.


QuentinGruber (commented):
"they started too again...?" Maybe the server rebooted and that's why

Nikolassparrow:
> "they started too again...?" Maybe the server rebooted and that's why

I don't think so, it was on wipe day only about 3-4 hours after the wipe had happened. I was also asking other groups and they told me their furnaces were acting funny, but would also randomly start working again too. When I tested this PR on singleplayer, I tried my best to stress test it with all types of fuel for the furnace, and it held up during the testing, so the issue could be unrelated to burnTime, but I'm not conclusive on that just yet.

QuentinGruber (commented):
Update on this? Cz i don't see how this PR fix something

QuentinGruber:
> > Update on this? Cz i don't see how this PR fix something
> 
> Furnaces are somewhat working on main servers, If people continue to report that furnaces are broken, I'll do some more further testing and open up another PR.

kk

PR #1874, opened 2024-01-31
Convey durability, randomize crowbar & convey durability, add boots functionality, skins and boots to default loadout (2016)

Nikolassparrow:
Accidentally messed up the other pull request cause of case sensitive branch names lol

https://www.youtube.com/watch?v=aFpSNGow3QI& (4:35 | 11:08)

Conveys used to have durability (5400, video above ^) and would get damaged based upon what MaterialType the player was walking on. That's still a WIP, but for now add durability to them and also add randomized durability to crowbars and conveys, which was a feature back in 2016 (also in the video above).

Also added boots functionality as they did nothing before (boots should also spawn in the world like conveys do), and added them to the players default loadout as players in 2016 spawned with the Gray Blue Work boots

Known bugs:
While testing, I noticed that boots and conveys don't change the sound of the players feet. There used to be a heavy boot sound when you would walk (like in real life), so that should be looked into.

Todo: Make boots have some kind of resistance to TrapEntities like they used to have in 2016.


src/servers/ZoneServer2016/entities/character.ts
```diff
@@ -151,6 +151,7 @@ export class Character2016 extends BaseFullCharacter {
   ) => void;
   timeouts: any;
   hasConveys: boolean = false;
+  hasBoots: boolean = false;
```
QuentinGruber, line 154: Since now there is hasConveys and hasBoots, refactoring to smth like `shoesType: ShoesTypeEnum` but this can be done later on

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -5433,6 +5433,10 @@ export class ZoneServer2016 extends EventEmitter {
     let durability: number = 2000;
     switch (true) {
       case this.isWeapon(itemDefinitionId):
+        if (itemDefinitionId == Items.WEAPON_CROWBAR) {
+          durability = Math.floor(Math.random() * 2001);
```
QuentinGruber, line 5437: Why 2001 ?

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -5441,6 +5445,9 @@ export class ZoneServer2016 extends EventEmitter {
       case this.isHelmet(itemDefinitionId):
         durability = 100;
         break;
+      case this.isConvey(itemDefinitionId):
+        durability = Math.floor(Math.random() * 5401);
```
QuentinGruber, line 5449: 5401?

QuentinGruber (approved):

src/servers/ZoneServer2016/entities/character.ts
Nikolassparrow: Yea for sure! I'll look into that once I get Conveys to lose durability in some way 👍

src/servers/ZoneServer2016/zoneserver.ts
Nikolassparrow: So the value 5400 or 2000 can be generated from the random number generator, but perhaps just doing 2000 and 5400 is the same thing?

src/servers/ZoneServer2016/zoneserver.ts
Nikolassparrow: ^^

src/servers/ZoneServer2016/zoneserver.ts
QuentinGruber: It's true that 1 is excluded so 0.999999999 * 2000 would be the max number but since you do Math.floor 1999.9999 would be 2000. Right now it could result in a crowbar having 2001 durability

Nikolassparrow: I see. I updated the code to be reflective of 5400 and 2000 :D

PR #1881, opened 2024-01-31
Deprecate furnace change

Nikolassparrow:
Okay, so several people are reporting that furnaces are still messed up, so for now just deprecate the recent furnace change to burnTime (keep smeltTime that shouldn't be causing issues) until more testing can be done.

### DIFF
--- a/src/servers/ZoneServer2016/managers/smeltingmanager.ts
+++ b/src/servers/ZoneServer2016/managers/smeltingmanager.ts
@@ -186,14 +186,6 @@ export class SmeltingManager {
     );
   }*/
 
-  getBurnTime(item: BaseItem): number {
-    if (item.itemDefinitionId == Items.CHARCOAL) {
-      return (this.burnTime = 2400000);
-    } else {
-      return (this.burnTime = 120000);
-    }
-  }
-
   private checkFuel(
     server: ZoneServer2016,
     entity: LootableConstructionEntity
@@ -202,7 +194,6 @@ export class SmeltingManager {
     for (const a in container!.items) {
       const item = container!.items[a];
       if (entity.subEntity!.allowedFuel.includes(item.itemDefinitionId)) {
-        this.getBurnTime(item);
         server.removeContainerItem(entity, item, entity.getContainer(), 1);
         if (item.itemDefinitionId == Items.WOOD_LOG) {
           // give charcoal if wood log was burned


QuentinGruber (approved):
I'm ok with it but i don't get how this would create the furnace bug

Nikolassparrow:
> I'm ok with it but i don't get how this would create the furnace bug

I'm not sure either... I just know before this update furnaces were working, and then I messed with them and now they aren't ):

QuentinGruber:
> > I'm ok with it but i don't get how this would create the furnace bug
> 
> I'm not sure either... I just know before this update furnaces were working, and then I messed with them and now they aren't ):

it's hard to track but the bug is probably more related to something like the timewrapper than your PR

Nikolassparrow:
> > > I'm ok with it but i don't get how this would create the furnace bug
> > 
> > 
> > I'm not sure either... I just know before this update furnaces were working, and then I messed with them and now they aren't ):
> 
> it's hard to track but the bug is probably more related to something like the timewrapper than your PR

You might be actually... I was talking to some people about it, and they told me their furnaces will work during the day, but turn off or not work when it turns night time. One person said: "i think you gotta burn wood for maybe 30 mins before putting the scrap in." Could be an issue with the recent time changes

QuentinGruber:
> > > > I'm ok with it but i don't get how this would create the furnace bug
> > > 
> > > 
> > > I'm not sure either... I just know before this update furnaces were working, and then I messed with them and now they aren't ):
> > 
> > 
> > it's hard to track but the bug is probably more related to something like the timewrapper than your PR
> 
> You might be right actually... I was talking to some people about it, and they told me their furnaces will work during the day, but turn off or not work when it turns night time. One person said: "i think you gotta burn wood for maybe 30 mins before putting the scrap in." Could be an issue with the recent time changes

ohhhh that could be linked to the nighttime being x2 + ingametime is reset every ingame day :D

Nikolassparrow:
> > > > > I'm ok with it but i don't get how this would create the furnace bug
> > > > 
> > > > 
> > > > I'm not sure either... I just know before this update furnaces were working, and then I messed with them and now they aren't ):
> > > 
> > > 
> > > it's hard to track but the bug is probably more related to something like the timewrapper than your PR
> > 
> > 
> > You might be right actually... I was talking to some people about it, and they told me their furnaces will work during the day, but turn off or not work when it turns night time. One person said: "i think you gotta burn wood for maybe 30 mins before putting the scrap in." Could be an issue with the recent time changes
> 
> ohhhh that could be linked to the nighttime being x2 + ingametime is reset every ingame day :D

For sure, hold out on pushing this update (very likely at this point the furnace issue is caused by something with in game time), my furnaces work perfectly during the day, but something happens at night that makes them screw up.

QuentinGruber:
well modifying the burntime property this way wasn't great

PR #1900, opened 2024-02-03
Fix for explosive damage inside secured shelters

Nikolassparrow:
Currently players can kill each other through walls with explosives, so add a check for a player being inside a secured shelter before damaging the player. Players can still be damaged through walls if the door isn't secured, so there'll have to be a future update for this, but for now this is a solid change.

### DIFF
--- a/src/servers/ZoneServer2016/zoneserver.ts
+++ b/src/servers/ZoneServer2016/zoneserver.ts
@@ -2309,7 +2309,7 @@ export class ZoneServer2016 extends EventEmitter {
             character.state.position,
             position,
             1.5
-          )
+          ) && !client?.character.isHidden
         ) {
           const distance = getDistance(position, character.state.position);
           const damage = 50000 / distance;
--- a/src/servers/ZoneServer2016/zoneserver.ts
+++ b/src/servers/ZoneServer2016/zoneserver.ts
@@ -2309,7 +2309,8 @@ export class ZoneServer2016 extends EventEmitter {
             character.state.position,
             position,
             1.5
-          ) && !client?.character.isHidden
+          ) &&
+          !client?.character.isHidden
         ) {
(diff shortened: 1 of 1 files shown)


Z1Meme:
This would unfortunately disable explosions from within a shelter from damaging players inside that shelter, so the new thing that people could do is set off an explosion from inside the shelter, killing raiders. The way this needs to be done is dropped items should have a isInside check when dropped and save the shelter that they're parented to. If an explosive and a player are inside together, it damages. If an explosive is outside and the player is inside, it wont.

Z1Meme:
> This would unfortunately disable explosions from within a shelter from damaging players inside that shelter, so the new thing that people could do is set off an explosion from inside the shelter, killing raiders. The way this needs to be done is dropped items should have a isInside check when dropped and save the shelter that they're parented to. If an explosive and a player are inside together, it damages. If an explosive is outside and the player is inside, it wont.

Having items parent to a secured shelter would also allow us to "despawn" dropped items in a shelter from outside players, as a form of continued ESP protection (similar to how workbenches, players, etc operate inside a shelter already).

Nikolassparrow:
> > This would unfortunately disable explosions from within a shelter from damaging players inside that shelter, so the new thing that people could do is set off an explosion from inside the shelter, killing raiders. The way this needs to be done is dropped items should have a isInside check when dropped and save the shelter that they're parented to. If an explosive and a player are inside together, it damages. If an explosive is outside and the player is inside, it wont.
> 
> Having items parent to a secured shelter would also allow us to "despawn" dropped items in a shelter from outside players, as a form of continued ESP protection (similar to how workbenches, players, etc operate inside a shelter already).

I see. I'll close this PR then for now as it would just be a small bandaid fix for a much larger issue (unless y'all still want to implement it otherwise :P)

PR #1906, opened 2024-02-04
bugfix/allWeaponsRandomDura

Nikolassparrow:
Was a slight oversight in #1899. Only crowbars (for now) should have random durability for weapons


QuentinGruber (approved):
Up to you but i found that actually cool

Nikolassparrow:
> Was it that bad ? actually found it cool that you could found almost broke Ar and stuff

I agree, weapons used to spawn with random durability back in the day, but I mainly made this PR to prevent things like flashlights from having durability 😂 

I'll update this PR to allow guns, hammers, and crowbars to the random durability list :D

Nikolassparrow:
Added a check for an item being crafted so crafted items don't get applied random durability, only item pickups.

QuentinGruber (changes requested):
Pr has render issue ( backpack  ) and circular issue in the vehicle object ( time ) which lead to a crash

QuentinGruber:
maybe not linked to this pr need to update it based on dev

QuentinGruber (changes requested):
Ok seems fist still have the issue

Nikolassparrow:
> Ok seems fist still have the issue

Wasn't able to re-produce the issue, how did you do it?

QuentinGruber:
> > Ok seems fist still have the issue
> 
> Wasn't able to re-produce the issue, how did you do it?

it was an old character my bad

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -5408,11 +5408,13 @@ export class ZoneServer2016 extends EventEmitter {
...
    * @param {number} itemDefinitionId - The itemDefinitionId of the item to generate.
    * @param {number} [count=1] - The count of the item.
+   * @param {number} [lastGeneratedTime=0] - The last generated time of the item.
    * @returns {BaseItem|undefined} The generated item, or undefined if the item definition is invalid.
    */
   generateItem(
     itemDefinitionId: number,
-    count: number = 1
+    count: number = 1,
+    lastGeneratedTime: number = 0
```
QuentinGruber, line 5417: maybe just changing lastGeneratedTime to a boolean would make the trick ?

QuentinGruber (approved):

PR #1182, opened 2022-12-24
Weighted lootspawn system

Z1Meme:
.

PR #1907, opened 2024-02-05
Fix containers not receiving entirety of loottable upon respawn

Nikolassparrow:
Found this bug while working on loot table stuff. I was extremely puzzled when I was doing /respawnloot and barely anything would spawn inside of the containers despite their spawnChances being 100. Essentially, the check for whether an item would spawn or not was going based upon the individual items weight, rather than the entirety of the lootTables spawnchance to spawn any specific item in said lootTable. (if this is wrong please correct me)


QuentinGruber:
Well it seems it was on purpose https://github.com/QuentinGruber/h1z1-server/pull/1182 so ask meme

Nikolassparrow:
> Well it seems it was on purpose #1182 so ask meme

There's no mention of containers in that PR. This is only an issue with lootableProp containers.

Z1Meme:
@Nikolassparrow Good catch

Z1Meme (approved):

PR #1952, opened 2024-02-21
AutoGroup

Nikolassparrow:
Been something that I've been thinking about for a while since re-logging/crashing is very common on H1EMU. You get kicked out of your group A LOT and it's annoying having to constantly invite the same people over and over again. 

This PR adds:

- Auto re-join group upon re-connecting to the server

Unfortunately, I don't know how to setup mongo so this hasn't been tested on a live server... I'm assuming it'd work, but definitely test this PR before merging.


Nikolassparrow:
> So `/group auto` and you rejoin your old group ?

Yes, /group auto will allow you to rejoin the group that you were in when you first initiated the disconnect.

Nikolassparrow:
Hey meme, could you verify that this PR is all good? I know you did a lot of the group related things, so I wanted to verify what I typed was correct

QuentinGruber (changes requested):
image

beside that it works but it's not a good behavior to have this in group view

Nikolassparrow:
> image
> 
> beside that it works but it's not a good behavior to have this in group view

I never had this occur while testing, how did you manage to get that message?

I do recall a wipe or two ago on the main servers I had that bug with "UNDEFINED - REPORTED THIS!", but it didn't break any group functionalities I believe..? Shouldn't be a thing though for sure.

QuentinGruber:
> > image
> > beside that it works but it's not a good behavior to have this in group view
> 
> I never had this occur while testing, how did you manage to get that message?
> 
> I do recall a wipe or two ago on the main servers I had that bug with "UNDEFINED - REPORTED THIS!", but it didn't break any group functionalities I believe..? Shouldn't be a thing though for sure.

it broke nothing but it's just there

Nikolassparrow:
> > > image
> > > beside that it works but it's not a good behavior to have this in group view
> > 
> > 
> > I never had this occur while testing, how did you manage to get that message?
> > I do recall a wipe or two ago on the main servers I had that bug with "UNDEFINED - REPORTED THIS!", but it didn't break any group functionalities I believe..? Shouldn't be a thing though for sure.
> 
> it broke nothing but it's just there

Okay so this led me down a rabbit hole of the groupManager class and basically how it all works from a bunch of testing and here's what I've found:

The groupId is determined by the nextGroupId which is default at 1. nextGroupId isn't saved by mongo so anytime the server crashes or restarts, the counter is set back down to 0. From testing this will cause MANY issues such as the users not being able to create/join groups because of a bug relating to how groupIds are detected. I believe this can be fixed by saving the nextGroupId to mongo and will thus allow the nextGroupId to infinitely increment without the fear of going back to 0, but more testing needs to be done to know for sure. 

 I wanted to take this a step further however and wished to have GroupSaveData to save group data upon a saveWorld or a server restart/crashing. What do you think about doing that? If so, there will need to be a lot more testing to ensure no strange bugs are related to saving GroupData. Meme might know more about what that could entail and would be nice to have his opinion on this

QuentinGruber:
> > > > image
> > > > beside that it works but it's not a good behavior to have this in group view
> > > 
> > > 
> > > I never had this occur while testing, how did you manage to get that message?
> > > I do recall a wipe or two ago on the main servers I had that bug with "UNDEFINED - REPORTED THIS!", but it didn't break any group functionalities I believe..? Shouldn't be a thing though for sure.
> > 
> > 
> > it broke nothing but it's just there
> 
> Okay so this led me down a rabbit hole of the groupManager class and basically how it all works from a bunch of testing and here's what I've found:
> 
> UNDEFINED - REPORT THIS happens when getClientByCharId returns undefined, so when the user is offline more than likely. This can be fixed by adding a getOfflineClientByCharId and searching mongo for the characterId (actively testing this new function)
> 
> The groupId is determined by the nextGroupId which is default at 1. nextGroupId isn't saved by mongo so anytime the server crashes or restarts, the counter is set back down to 0. From testing this will cause MANY issues such as the users not being able to create/join groups because of a bug relating to how groupIds are detected. I believe this can be fixed by saving the nextGroupId to mongo and will thus allow the nextGroupId to infinitely increment without the fear of going back to 0, but more testing needs to be done to know for sure.
> 
> I wanted to take this a step further however and wish to have a GroupSaveData object to save group data upon a saveWorld or a server restart/crashing. What do you think about doing that? If so, there will need to be more testing to ensure no strange bugs are related to saving GroupData. Meme might know more about what that could entail and would be nice to have his opinion on this

Ok so my take is => We abandon this PR , create an issue about saving group data to mongo and work on it. Would be better than duck tape, and no need to include the group save data in world save data we will directly update and fetch stuff on mongo. I could setup some of this

Nikolassparrow:
> beside that it works but it's not a good behavior to have this in group view
> 
> I never had this occur while testing, how did you manage to get that message? I do recall a wipe or two ago on the main servers I had that bug with "UNDEFINED - REPORTED THIS!", but it didn't break any group functionalities I believe..? Shouldn't be a thing though for sure.
> 
> it broke nothing but it's just there
> 
> Okay so this led me down a rabbit hole of the groupManager class and basically how it all works from a bunch of testing and here's what I've found:
> 
> UNDEFINED - REPORT THIS happens when getClientByCharId returns undefined, so when the user is offline more than likely. This can be fixed by adding a getOfflineClientByCharId and searching mongo for the characterId (actively testing this new function)
> 
> The groupId is determined by the nextGroupId which is default at 1. nextGroupId isn't saved by mongo so anytime the server crashes or restarts, the counter is set back down to 0. From testing this will cause MANY issues such as the users not being able to create/join groups because of a bug relating to how groupIds are detected. I believe this can be fixed by saving the nextGroupId to mongo and will thus allow the nextGroupId to infinitely increment without the fear of going back to 0, but more testing needs to be done to know for sure.
> 
> I wanted to take this a step further however and wish to have a GroupSaveData object to save group data upon a saveWorld or a server restart/crashing. What do you think about doing that? If so, there will need to be more testing to ensure no strange bugs are related to saving GroupData. Meme might know more about what that could entail and would be nice to have his opinion on this
> 
> Ok so my take is => We abandon this PR , create an issue about saving group data to mongo and work on it. Would be better than duck tape, and no need to include the group save data in world save data we will directly update and fetch stuff on mongo. I could setup some of this

Okay sounds good 👍 I'll close this PR for now but keep the branch so I can see the logic for later 👍

PR #1957, opened 2024-02-24
Bandaid vehicle fairplay fix

Nikolassparrow:
Fairplay kicking in vehicles has something to do with the way time works in the distance equation inside of checkVehicleSpeed (fairplaymanager)

This is just a bandaid fix until a more permanent one can be found.


QuentinGruber (commented):
Idk what @avcio9 think of this if you want maybe it's better to disable this until we fix it

avcio9:
Are we using a main or dev branch on main servers? the fix i made a few days ago is on dev but not merged into main branch

avcio9:
Looks like i've found the issue, it was hard to catch on local server cause of low ping

QuentinGruber:
> Are we using a main or dev branch on main servers? the fix i made a few days ago is on dev but not merged into main branch

We use the latest dev i will push a regular version today

QuentinGruber:
> Looks like i've found the issue, it was hard to catch on local server cause of low ping

Make sense gj!

QuentinGruber (approved):

PR #1960, opened 2024-02-25
Barbed wire changes

Nikolassparrow:
In #1765, I didn't realize that the crafted barbed wire type was the one which had the wooden posts on the side of them that also acted as an object with boundaries to prevent players from walking through them like with the other model.

This PR does:

- Changes the craftable barbed wire model type to the correct type of barbed wire with wooden posts
- Adds meleeing, shooting and damaging to TrapEntities,
- Adds functionality to barbed wire with wooden posts around the map
- Fixed issue where TrapEntities weren't getting recognized by CommandInteractionString
- Type safety inside of the TrapEntity class


QuentinGruber:
isn't it too easy to destroy ? like 7 bullet of ar ?

data/2016/zoneData/Z1_taskProps.json
```diff
@@ -19509,5 +19508,7635 @@
...
     ]
+  },
+  {
+    "actor_file": "Common_Props_BarbedWire_Posts.adr",
+    "modelId": 43,
+    "renderDistance": 50,
+    "instances": [
+      {
+        "position": [
+          -3479.26318359375,
```
QuentinGruber, line 19519: these spawned barbedWire aren't destroyed when "killed" the model still there but make no damage

src/servers/ZoneServer2016/entities/trapentity.ts
```diff
@@ -243,4 +245,31 @@ export class TrapEntity extends BaseSimpleNpc {
         return isInsideCube(Array.from(position) as Point3D, this.cubebounds);
     }
   }
+
+  OnProjectileHit(server: ZoneServer2016, damageInfo: DamageInfo) {
+    const damage = damageInfo.damage * 6; // bullets do more to damage traps
```
QuentinGruber, line 250: should do less imo

QuentinGruber (approved):

data/2016/zoneData/Z1_taskProps.json
Nikolassparrow: I remember this problem from #1765, this is also an issue with punji sticks and has been since they were implemented. 

A fix for this could be to check if a barbed wire is a world spawn or a player placed it. From this check we can make it so if it's a world spawned, then to not allow it to be destroyed, and if it's a player spawn then to allow anyone to destroy it. If I recall correctly I think this is how DBG did it also 😂😂

src/servers/ZoneServer2016/entities/trapentity.ts
Nikolassparrow: You reckon so? What would be a good damage multiplier?

QuentinGruber: idk actually lets try with this and see players reactions

PR #1973, opened 2024-02-27
Add restricted building rulesets

Nikolassparrow:
So something that I remembered from 2016 was the "restricted building" ruleset image on the server select screen that basically told you if a server allowed you to build in pois or not. I miss the hammer and triangle picture so this is something I've tried to work on for a while but finally managed to get working 😂 I found it from decompiling the CharacterSelect.gfx actionscript file, there were some other gamemodes included in there too!

So this PR: 

- Adds restricted building ruleset & image to server select screen
- Add isNoBuildInPois config variable for building in poi and removed allowPOIPlacement

### DIFF
--- a/data/2016/sampleData/defaultconfig.yaml
+++ b/data/2016/sampleData/defaultconfig.yaml
@@ -12,6 +12,7 @@ server:
   isPvE: false
   isHeadshotOnly: false
   isFirstPersonOnly: false
+  isNoBuildInPois: true
   baseConstructionDamage: 34000
 
 # Fairplay / anticheat config
@@ -102,7 +103,6 @@ speedtree:
   maxTreeHits: 20 # maximum hits it takes to chop a tree
 
 construction:
-  allowPOIPlacement: false
   allowStackedPlacement: false
   allowOutOfBoundsPlacement: false
   placementRange: 30
--- a/data/2016/sampleData/defaultconfig.yaml
+++ b/data/2016/sampleData/defaultconfig.yaml
@@ -103,6 +103,7 @@ speedtree:
   maxTreeHits: 20 # maximum hits it takes to chop a tree
 
 construction:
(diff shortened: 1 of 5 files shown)


data/2016/sampleData/defaultconfig.yaml
```diff
@@ -102,7 +103,6 @@ speedtree:
   maxTreeHits: 20 # maximum hits it takes to chop a tree
 
 construction:
-  allowPOIPlacement: false
```
QuentinGruber, line 105: maybe just comment it and add in the comment that isNoBuildInPois exist now

src/servers/ZoneServer2016/managers/constructionmanager.ts
```diff
@@ -347,13 +346,14 @@ export class ConstructionManager {
...
   detectPOIPlacement(
+    server: ZoneServer2016,
     itemDefinitionId: number,
     position: Float32Array,
     client: Client,
     isInsidePermissionedFoundation: boolean
   ): boolean {
     if (client.isDebugMode) return false;
-    if (this.allowPOIPlacement) return false;
+    if (server.isNoBuildInPois) return false;
```
QuentinGruber, line 356: wouldbe better to do this directly when detectPOIPlacement method is called

QuentinGruber (changes requested):

src/servers/ZoneServer2016/managers/constructionmanager.ts
Nikolassparrow: Done 👍

data/2016/sampleData/defaultconfig.yaml
Nikolassparrow: Done 👍

QuentinGruber (approved):

PR #1987, opened 2024-03-01
Hammer cooldown for repairing constructions

Nikolassparrow:
https://www.youtube.com/watch?v=WpSL7cDjQRs&t=429s (3:05, other parts of the vids he shows him hammering too)

So currently, a common thing players like to do when getting raided is to have multiple people melee spamming hammers on the base so explosives like ethanol/biofuel have less of a chance to destroy the construction - not a problem with IEDs b/c you can light multiple of them at once (the math to determine if a construction will be repaired or not while in the middle of exploding is completely random, it's based upon timing, server lag, whole bunch of other shit etc.) 

Anyways, this is super boring gameplay to just have people sitting there spamming to repair the base than actually defending it... So this PR adds a 1 second delay between repair times to prevent this kind of gameplay.

I also moved the logic pertaining to exploding nearby explosives up to speed up the exploding times of eth/bio by a small amount.

### DIFF
--- a/src/servers/ZoneServer2016/zoneserver.ts
+++ b/src/servers/ZoneServer2016/zoneserver.ts
@@ -2306,6 +2306,23 @@ export class ZoneServer2016 extends EventEmitter {
 
     const sourceEntity = this.getEntity(npcTriggered),
       sourceIsVehicle = sourceEntity instanceof Vehicle2016;
+
+    // explode nearby explosives first
+    for (const explosive in this._explosives) {
+      const explosiveObj = this._explosives[explosive];
+      if (explosiveObj.characterId != npcTriggered) {
+        if (getDistance(position, explosiveObj.state.position) < 2) {
+          await scheduler.wait(100);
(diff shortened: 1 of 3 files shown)


src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -2306,6 +2306,23 @@ export class ZoneServer2016 extends EventEmitter {
 
     const sourceEntity = this.getEntity(npcTriggered),
       sourceIsVehicle = sourceEntity instanceof Vehicle2016;
+
+    // explode nearby explosives first
```
QuentinGruber, line 2310: but doesn't it allow to kill with an ied in PVE ?

QuentinGruber (changes requested):

src/servers/ZoneServer2016/zoneserver.ts
Nikolassparrow: No sir 👍 All is good for the PvE-er's

QuentinGruber (approved):

Issue #1979, opened 2024-03-01
Allow players to waste fuel to refuel up to 100% 


QuentinGruber:
ok split this in another pr if you want this do be merge since it allow players to waste fuel while being at 100%

_Originally posted by @QuentinGruber in https://github.com/QuentinGruber/h1z1-server/pull/1972#discussion_r1507369357_

PR #1996, opened 2024-03-02
Allow refueling to 100, but not exceed

Nikolassparrow:
#1979, when the resourceids.fuel + fuelValue is over or equal to 10000, adjust the fuel level to be 10000 to prevent the fuel resource from being broken

### DIFF
--- a/src/servers/ZoneServer2016/zoneserver.ts
+++ b/src/servers/ZoneServer2016/zoneserver.ts
@@ -6791,12 +6791,6 @@ export class ZoneServer2016 extends EventEmitter {
       );
       return;
     }
-    const vehicle = this._vehicles[vehicleGuid];
-    if (vehicle._resources[ResourceIds.FUEL] + fuelValue > 10000) {
-      // Prevent players from overfilling their car to reserve fuel.
-      this.sendAlert(client, "Fuel level is not low enough to refuel");
-      return;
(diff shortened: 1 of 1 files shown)


src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -6791,12 +6791,6 @@ export class ZoneServer2016 extends EventEmitter {
       );
       return;
     }
-    const vehicle = this._vehicles[vehicleGuid];
-    if (vehicle._resources[ResourceIds.FUEL] + fuelValue > 10000) {
-      // Prevent players from overfilling their car to reserve fuel.
-      this.sendAlert(client, "Fuel level is not low enough to refuel");
-      return;
```
QuentinGruber, line 6798: Yep but by removing that players can now waste fuel trying to refuel while being at 100% the operation should fail in this case

QuentinGruber (changes requested):

src/servers/ZoneServer2016/zoneserver.ts
Nikolassparrow: Ahh good point, Idk why I removed that check looking back 😂😂 Should be good to go

QuentinGruber (approved):

Issue #1990, opened 2024-03-02
need durability switches modified in zoneServer, baseFullCharacter and Character classes


QuentinGruber:
Still need durability switches modified in zoneServer, baseFullCharacter and Character classes,
Additionally we should remove this item from default loadout or increase the durability loss to like 500

_Originally posted by @avcio9 in https://github.com/QuentinGruber/h1z1-server/issues/1933#issuecomment-1958879026_

Nikolassparrow:
> > Still need durability switches modified in zoneServer, baseFullCharacter and Character classes, Additionally we should remove this item from default loadout or increase the durability loss to like 500
> 
> No need for durability switches, I just changed the ITEM_TYPE on the Jagged Skinning Knife to 20 so it'll get thrown into the isWeapon function! (Hopefully doesn't break anything lol)
> 
> And 500 damage to the knife? That's quite a lot and would only let you harvest 4 npcs.. How about 250 or 400 so the player can at least have some uses with the item? Skinning knifes are also a bit expensive to craft (3 scrap 1 wood stick)

There won't need to be additional durability switches as I changed the ITEM_TYPE for the jagged skinning knife to match the isWeapon functions ITEM_TYPE, so anytime isWeapon is called, the jagging skinning knife will be considered into that equation

PR #1998, opened 2024-03-02
Fix skinning knife durability/damage

Nikolassparrow:
Fix for #1995 and #1990, I'm pretty sure changing the ITEM_TYPE on an item will cause crashes for some odd reason (possible built in H1Z1 function that checks for validity or something). Should be good now though 👍


QuentinGruber (approved):
tested ingame works great

PR #1972, opened 2024-02-27
Make wrench consistent with July 2016 update

Nikolassparrow:
2bcb2c526fd74a3050d7bd99166ca96c

In the July 7th, 2016 update facepunch added Anti-Vehicle hoarding measures to prevent players from excessively storing cars.  Currently on H1EMU, wrenches have 2k durability and repair the vehicle 5% per swing for a maximum of 100%. In the Anti-Hoarding update, wrenches had 2.5k durability and repaired the vehicle by 2% per swing for a maximum of 50%. This PR is to make the wrench more in-line with that update.

This PR changes - 

- Modified wrench to have 2.5k durability and repair a vehicle by 2% per swing
- Fixed a bug where you couldn't refuel a vehicle if the fuel amount was between 76-99%.

PR #1999, opened 2024-03-03
Lower vehicle decay rate

Nikolassparrow:
So to follow-up with #1972, in the same screenshot DBG talks about how they lowered vehicle decay to quote: "decay their condition slowly over time. they [vehicles] will fully decay in a little over two days." 

This PR lowers the baseVehicleDamage to be 2000 which is a little over two days (50 hours to be exact). 

In my personal opinion too, I think vehicle decay was a little too quick before also. It was around ~15 hours even if you only had 1 car and it was full hp! Also I was wondering to maybe buff maxVehiclesPerArea by 1? If I'm not mistaken, in 2016 you could have 3 cars but once you had 4 then the vehicle decay would ramp up, but I could be remembering wrong


QuentinGruber (commented):
awaiting for Avcio approval

QuentinGruber (approved):

PR #2001, opened 2024-03-04
Fix isNoBuildInPOIs boolean algebra

Nikolassparrow:
Title ^ my dumbass accidentally did if (server.isNoBuildInPois) instead of (!server.isnoBuildInPOIs) 😭

While I was at it I moved the !server.isNoBuildInPois into the detectPoiPlacement where allowPoiPlacement was originally

### DIFF
--- a/src/servers/ZoneServer2016/managers/constructionmanager.ts
+++ b/src/servers/ZoneServer2016/managers/constructionmanager.ts
@@ -352,13 +352,16 @@ export class ConstructionManager {
     return false;
   }
   detectPOIPlacement(
+    server: ZoneServer2016,
     itemDefinitionId: number,
     position: Float32Array,
     client: Client,
     isInsidePermissionedFoundation: boolean
   ): boolean {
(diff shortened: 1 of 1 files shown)


src/servers/ZoneServer2016/managers/constructionmanager.ts
```diff
@@ -352,13 +352,16 @@ export class ConstructionManager {
     return false;
   }
   detectPOIPlacement(
+    server: ZoneServer2016,
```
QuentinGruber, line 355: idk why you keep adding the whole server as an argument when you can just ignore the function call based on your boolean

Nikolassparrow: It was to provide clarity for people learning. You look for the isNoBuildInPoi variable and see it's not even in the detectPOIPlacement when you would expect it to be. Is it gonna completely break the code for using the server instance or something 😂😂?

PR #2003, opened 2024-03-07
Candles

Nikolassparrow:
Iniko on discord suggested that candles get added so that's what this PR is. Added placement of candles to placeTemporaryEntity and set it to 1 hr which is what it was back in the day.


QuentinGruber (approved):
nice

PR #2006, opened 2024-03-09
Bugfix/EmptyBottleBeeBox

Nikolassparrow:
Was reported by clever_reeeeee in the H1EMU discord. 

Empty water bottles were getting removed from containers despite having no honeycomb inside them at the time of removal, so this PR just adds a check for if the current item is an empty water bottle and if the container only contains that water bottle (container.items.length = 1) then continue out of the loop.

### DIFF
--- a/src/servers/ZoneServer2016/managers/smeltingmanager.ts
+++ b/src/servers/ZoneServer2016/managers/smeltingmanager.ts
@@ -268,6 +268,9 @@ export class SmeltingManager {
   ) {
     for (const a in container.items) {
       const item = container.items[a];
+      if (item.itemDefinitionId == Items.WATER_EMPTY 
+        && Object.keys(container.items).length == 1) 
+        continue;
       if (item.itemDefinitionId != Items.WATER_EMPTY) continue;
       if (subEntity.currentTicks >= subEntity.requiredTicks) {
         subEntity.currentTicks = 0;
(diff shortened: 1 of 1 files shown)


src/servers/ZoneServer2016/managers/smeltingmanager.ts
```diff
@@ -268,6 +268,11 @@ export class SmeltingManager {
   ) {
     for (const a in container.items) {
       const item = container.items[a];
+      if (
+        item.itemDefinitionId == Items.WATER_EMPTY &&
+        Object.keys(container.items).length == 1
+      )
+        continue;
       if (item.itemDefinitionId != Items.WATER_EMPTY) continue;
```
QuentinGruber, line 276: doesn't it conflict with this statement?, currently it will not continue the loop only if there is a WATER_EMPTY but the items len isn't 1

Nikolassparrow: Should be fine, I was kind of in a rush when I made this PR and I was thinking earlier that there could be a problem if an empty water bottle + wax are the only things in the bee box (this could happen naturally from the honeycomb turning into honey) and that could cause the bug again, so I'm gonna test this PR some more when I get an opportunity (tonight or tomorrow). 👍

Nikolassparrow: Okay upon further investigation 🔍I fixed this issue I referenced above ^ I added a check to see if the current item is an empty water bottle and that the container is either empty or currently not containing honeycomb, in that case then to exit the scope of the for loop (container.items). Like I mentioned above, it shouldn't interfere with the next line of code (itemdefId != water_empty), from the tests I've done, dew collectors and bee boxes both worked as intended.

QuentinGruber (approved):

PR #2007, opened 2024-03-10
Fix melee damage while respawning

Nikolassparrow:
Someone did this to me on wipe, super cringe stuff (sadly AMD didn't clip it, I wish it did). Essentially while a player is respawning, you can melee their corpse and when the player respawns, they will be low hp or nearly dead.. So this PR fixes that by adding checks to the OnMeleeHit and Damage functions in the character class to see if a player is currently respawning

### DIFF
--- a/src/servers/ZoneServer2016/entities/character.ts
+++ b/src/servers/ZoneServer2016/entities/character.ts
@@ -1047,7 +1047,8 @@ export class Character2016 extends BaseFullCharacter {
       oldHealth = this._resources[ResourceIds.HEALTH];
     if (!client) return;
 
-    if (this.isGodMode() || !this.isAlive || this.isRespawning || damage <= 25) return;
+    if (this.isGodMode() || !this.isAlive || this.isRespawning || damage <= 25)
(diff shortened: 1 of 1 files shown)


src/servers/ZoneServer2016/entities/character.ts
```diff
@@ -1038,6 +1038,7 @@ export class Character2016 extends BaseFullCharacter {
   async damage(server: ZoneServer2016, damageInfo: DamageInfo) {
     if (
       server.isPvE &&
+      this.isRespawning &&
```
QuentinGruber, line 1041: bad logic it's why the test fail, try running "npm t" to see the issue it create

QuentinGruber (changes requested):

src/servers/ZoneServer2016/entities/character.ts
Nikolassparrow: Ahh okay 👍

QuentinGruber (approved):

PR #2047, opened 2024-04-03
Account Inventory and more WIP

jasonmoonen:
This PR includes the following changes and additions:

- Send Z1BR rulesets.
- Only send item definitions for items in the player's inventory.
- You can now salvage ammo in bulk.
- Proximity will now update when you drop an item.
- You should be able to trade with people; this means your shots should still count when you just died.
- Account inventory is now functional.
- Added account crates and bags. There are only a few working; if you want to know which ones work, you can look at RewardCrates.json.
- Updated the vehicle kit.
- Flash and fire traps are now functional.
- Landmines and traps will now be activated 10 seconds after being placed.
- You can now harvest a maximum of one brain from a zombie.
- You can now harvest cloth from screamers as well.
- You can no longer equip grenades in weapon slots, except for smoke grenades.
- Smoke grenades are now functional.
- Fixed the dam location for the /tp command.
- Fixed an exploit with stashes.
- Placement of stashes inside or near bases is now blocked.
- You can no longer place stashes near each other.
- Limited stash storage to 5000 bulk or 10 slots.
- Kits will now give you full durability conveys.


src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -6303,6 +6502,34 @@ export class ZoneServer2016 extends EventEmitter {
...
+
+    console.log(data);
+
+    if (!client) {
+      this.sendDataToAll("Command.DeliveryManagerStatus", data);
+      console.log("a");
+      return;
+    }
+
+    console.log("cl");
```
QuentinGruber, line 6529:
```suggestion
```

QuentinGruber (approved):

PR #2053, opened 2024-05-02
Save groups mongo

### LINKED ISSUE #1984: Save groups data to mongo
#1952 comment: Ok so my take is => create an issue about saving group data to mongo and work on it. Would be better than duck tape, and no need to include the group save data in world save data we will directly update and fetch stuff on mongo. I could setup some of this

QuentinGruber:
Close #1984


src/servers/ZoneServer2016/managers/groupmanager.ts
```diff
@@ -124,65 +120,87 @@ export class GroupManager {
...
 
-  createGroup(server: ZoneServer2016, leader: Client) {
-    const groupId = this.nextGroupId;
-    this.groups[groupId] = {
-      groupId: groupId,
+  async createGroup(server: ZoneServer2016, leader: Client) {
+    // Since groupId is a Number and Numbers can't store too large number we have to do this little trick
+    const maxSafeBigInt = BigInt(Number.MAX_SAFE_INTEGER);
+    const groupId: number = Number(
+      BigInt(generate_random_guid()) % maxSafeBigInt
```
zephykasmar, line 162: What's the need of generating a random GUID to generate a random number? Why not something like `const groupId: number = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER)`.

Then we won't need this trick and import.

QuentinGruber: Hey, true idk why i messed my mind on this :sweat_smile: thx !

PR #2074, opened 2024-05-27
Rework Account Inventory

QuentinGruber:
- [x] Migrate old inventories at server start ( will require a wipe i abandon the idea of a migration )
- [x] issues with itemguids ? Why using them when you can use itemdef? collision for sure


jasonmoonen:
> * [ ]  issues with itemguids ? Why using them when you can use itemdef? collision for sure

Z1BR uses incremental IDs. Instead of generating new IDs, I opted to use GUIDs instead. See the screenshot below (item GUIDs are hidden for privacy reasons):
image

jasonmoonen (approved):

PR #2109, opened 2024-07-21
feature/AI

QuentinGruber:
original avcio's work on today version


QuentinGruber:
doesn't seems to work under windows :/ weird but probably fixable since it's a rust module using the napi that is compiled

PR #2119, opened 2024-08-06
optimiseSpeedTreeManager

QuentinGruber:
- **rename speedtree property to speedtreedestroyed**
- **speedtreelist as a map**
- **remove allowedTrees for constructions from Z1_speedTrees.json**
- **use foreach from map**
- **remove treeId field**


QuentinGruber:
it remove almost 100mb from the heap and speedup start time and placement. ( funny enough unit test pass in 2.2 sec when it was 3.6 sec previously )

PR #2122, opened 2024-08-14
For our dear friend

jasonmoonen:
Added a gravestone with some goodies, the stash should only spawn every wipe.


QuentinGruber (approved):
Will merge and release that on next week wipe

PR #2144, opened 2024-09-25
Minor changes

jasonmoonen:
1. Restored the population level.
2. Players inside a shelter should no longer receive damage, unless the door is open, of course.

### DIFF
--- a/src/packets/LoginUdp/LoginUdp_11/loginpackets.ts
+++ b/src/packets/LoginUdp/LoginUdp_11/loginpackets.ts
@@ -28,9 +28,10 @@ const serverField: any[] = [
   { name: "descriptionId", type: "uint32" },
   { name: "reqFeatureId", type: "uint32" },
   { name: "serverInfo", type: "string" },
+  { name: "populationLevel", type: "uint32", defaultValue: 0 },
   // The "populationNumber" dword was split into 2 uint16s to show server population numbers on the main menu -Meme
-  { name: "populationNumber", type: "uint16", defaultValue: 0 },
-  { name: "maxPopulationNumber", type: "uint16", defaultValue: 1 },
+  //{ name: "populationNumber", type: "uint16", defaultValue: 0 },
+  //{ name: "maxPopulationNumber", type: "uint16", defaultValue: 1 },
   { name: "populationData", type: "string" },
   { name: "AccessExpression", type: "string", defaultValue: "" },
   { name: "allowedAccess", type: "boolean" }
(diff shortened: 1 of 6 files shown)


src/servers/ZoneServer2016/handlers/commands/commands.ts
```diff
@@ -158,7 +160,7 @@ export const commands: Array<Command> = [
   },
   {
     name: "pop",
-    permissionLevel: PermissionLevels.DEFAULT,
+    permissionLevel: PermissionLevels.ADMIN,
```
QuentinGruber, line 163: Isn't it a bit much? Like i guess you will remove every logs that display world pop? I'm ok for the main menu but that's all were not gonna do stuff like that to give the illusion there is players on the server

QuentinGruber (changes requested):

src/servers/ZoneServer2016/handlers/commands/commands.ts
QuentinGruber: A better idea could be do display other data too not only the actual player count but maybe the 24h connections ?

QuentinGruber: > A better idea could be do display other data too not only the actual player count but maybe the 24h connections ?

http://charts.h1emu.com/Last%20month%20connections.html

LegendsNeveerrDie: Even with the changes they can view the POP with Charts and the Website. Thats all thats needed. 

We are not removing it fully. Put in mind again daybreak had No charts, no pop command, no website display and servers didn't have the pop shown until 2018 as they never had that many players.

Players have told me they hated the POP count being shown too from day one, players would use it every 20mins to an hour. Im going on player feedback and I asked JSON to push this for me. :) 

Looking back at Jan 2023's update display was good without them.

LegendsNeveerrDie (approved):
Im with the feedback on this one. Its good changes to help the servers POP. :)

src/packets/LoginUdp/LoginUdp_11/loginpackets.ts
```diff
@@ -28,9 +28,10 @@ const serverField: any[] = [
   { name: "descriptionId", type: "uint32" },
   { name: "reqFeatureId", type: "uint32" },
   { name: "serverInfo", type: "string" },
+  { name: "populationLevel", type: "uint32", defaultValue: 0 },
```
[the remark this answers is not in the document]
jasonmoonen: I've already asked Eavilaswayce to update Asset_256

QuentinGruber (approved):

src/servers/ZoneServer2016/handlers/commands/commands.ts
QuentinGruber: ok sounds good to me

PR #2153, opened 2024-09-28
Handle skins correctly

jasonmoonen:
H1Z1 v0 195 4 147586 id=596c7c7e 10_16_2024 9_04_47 PM
H1Z1 v0 195 4 147586 id=2903fbc2 10_2_2024 2_54_56 PM


QuentinGruber:
This is the way they do it in z1br? Like can't the client use the local files it has?

jasonmoonen:
> This is the way they do it in z1br? Like can't the client use the local files it has?

Yes we can put it in the assets, but this is the way Daybreak always did it

QuentinGruber:
Ok well since this file is so huge plz minify it it's almost the git limit it will also help nodejs parsing it

Z1Meme (approved):

src/servers/ZoneServer2016/managers/worlddatamanager.ts
```diff
@@ -1155,7 +1156,8 @@ export class WorldDataManager {
         new Float32Array(entityData.rotation),
         server,
         getCurrentServerTimeWrapper().getTruncatedU32(),
-        entityData.vehicleId
+        entityData.vehicleId,
+        entityData?.shaderGroupId ?? 0
```
QuentinGruber, line 1160: Why "?" on entitydata?

src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -3737,6 +3737,9 @@ export class ZonePacketHandlers {
       case "Group.Leave":
         this.leaveGroup(server, client);
         break;
+      case "Ping":
```
QuentinGruber, line 3740: It's a real packet or crafted from us?

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -1880,6 +1876,17 @@ export class ZoneServer2016 extends EventEmitter {
   }
 
   async sendInitData(client: Client) {
+    // Load character before sending sendSelf, as we need to send item definitions for the inventory to the client first. This is the same behaviour as Z1BR
+    await this.fetchCharacterData(client);
+
+    // This packet is only necessary for transitioning between different zones.
```
QuentinGruber, line 1882: Quick question unrelated, do you know how to switch zones?

QuentinGruber (changes requested):
Impressive ! I just lock this Pr since i guess it's not compatible with old databases and so we have to await the wipe

src/servers/ZoneServer2016/managers/worlddatamanager.ts
jasonmoonen: In case it doesn't exist, though probably unnecessary.

src/servers/ZoneServer2016/zonepackethandlers.ts
jasonmoonen: No, a real packet from KOTK. Just guessing the name and structure though.

src/servers/ZoneServer2016/zoneserver.ts
jasonmoonen: Yes, it's extensively used in BR

QuentinGruber:
@jasonmoonen skins seems to be fixed only for the player skins aren't replicated

QuentinGruber (approved):

PR #2166, opened 2024-10-01
fix_unoptimized_code

QuentinGruber:


resinate:
i did some checking
      client.character.state.position = packet.data.position; = black screen when relog
      the packet only sends 3 values NOT 4

QuentinGruber:
Yep we saw that it's fixed now

resinate:
this.state.position = new Float32Array([0,0,0]); has to be 4 0s kept getting stupid black screen again LOL

PR #2189, opened 2024-10-23
Some more fixes

jasonmoonen:
1. Fixes group kick from UI
2. Send base parts before players are sent to the clients
3. Increase range for base parts so players are hidden behind them
4. You can now specify a player name in the parachute command


avcio9:
Do we want to see the whispers?

src/servers/ZoneServer2016/handlers/commands/commands.ts
```diff
@@ -414,6 +414,11 @@ export const commands: Array<Command> = [
         targetClient,
         `[Whisper from ${client.character.name}]: ${message}`
       );
+      if (!client.isAdmin) {
+        server.sendChatTextToAdmins(
+          `[${client.character.name} whispered to ${targetClient.character.name}]: ${message}`
```
QuentinGruber, line 419: .

QuentinGruber (changes requested):

QuentinGruber (approved):

PR #2204, opened 2024-10-27
fixed infinite armor bug

Z1Meme:


src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -5846,12 +5846,8 @@ export class ZoneServer2016 extends EventEmitter {
    */
   isArmor(itemDefinitionId: number): boolean {
     return (
-      this.getItemDefinition(itemDefinitionId)?.DESCRIPTION_ID ==
-        StringIds.TACTICAL_ARMOR ||
-      this.getItemDefinition(itemDefinitionId)?.DESCRIPTION_ID ==
-        StringIds.WOODEN_ARMOR ||
-      this.getItemDefinition(itemDefinitionId)?.DESCRIPTION_ID ==
-        StringIds.PLATED_ARMOR
+      this.getItemDefinition(itemDefinitionId)?.ITEM_CLASS ==
```
jasonmoonen, line 5849: Are only body armors using this item class? If not we might need to use the IS_ARMOR bool as well

jasonmoonen (changes requested):

jasonmoonen (approved):

PR #2206, opened 2024-10-27
remove assets hashs from db

QuentinGruber:
- **rm ASSET_HASHES from db**
- **rm workaround**


QuentinGruber:
need to test it ingame

PR #2226, opened 2024-10-31
Reorganize how entities are stored in zoneserver

Z1Meme:


Z1Meme:
DO NOT MERGE until feature/BWC and any other major PRs are merged

PR #2243, opened 2024-11-12
traps ai

QuentinGruber:
- **start moving traps to h1emu-ai**
- **arm traps**


QuentinGruber:
We must have a client side way of doing that, so for now even tho this pr work it will be closed

PR #2260, opened 2024-11-28
world save fix? 🙏

avcio9:


QuentinGruber:
The issue with that is that a crash at the wrong time make us loose the whole db

Issue #2246, opened 2024-11-18
.308 Noscope nerf


Z1Meme:
Need to nerf .308 noscope accuracy as its apparently more accurate than scoping in

johnoclockdk:
why is this not close yet

QuentinGruber:
> why is this not close yet

mistake, thx !

PR #2289, opened 2024-12-14
Fix 308 no scope

Nikolassparrow:
Reported by: anyone that has ever pvped on H1EMU before XD (Fix for #2246)

No seriously, this has been a long requested feature as currently players majority of the time resort to no scoping as it was 100% accurate. 

These numbers could be tweaked, I fiddled around with a bit with these and they felt the best, but in my opinion, this should be tested on Deadzone to get players opinions before being pushed to live. The inaccuracy while no scoping is big enough to be noticeable, but not big enough to where no scoping is impossible.

Video: https://streamable.com/gykq1v


Ultrax1995:
good change mate, I testes it with 15 people on DeadZone.

Nikolassparrow:
> good change mate, I testes it with 15 people on DeadZone.

Hell yeah 👍🏼
Thanks for the update

QuentinGruber (approved):
Nice it will be released tomorrow, thx ! 🤠

PR #2290, opened 2024-12-14
Increase gunpowder bulk 

Nikolassparrow:
Reported by: Multiple teams

Currently the meta on H1EMU is to log out with loot, specifically gunpowder and explosive items, as it's impossible to have lose it. From what I've heard, teams tend to log out with most, if not all their gunpowder making it a luxury, rather than common place during a raid. It's also kind of boring to know people will likely log out with valuable loot if they feel they won't win. 

I wasn't sure what to increase the bulk amount too, but I've heard from players that it should, at minimum, be the same as fertilizer (50), so I started with that. I've also heard it should be 100 but that seems a bit high?


QuentinGruber (approved):

QuentinGruber:
### Merge activity

* **Dec 15, 9:31 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/QuentinGruber/h1z1-server/2290).

PR #2295, opened 2024-12-16
Bugfix/locked vehicle on death

Nikolassparrow:
Reported by: KoreaNTC (Fix for #2296)

So when you kill the last player in a vehicle, the vehicle will remain locked until the killed player respawns. Currently players are abusing this bug by remaining on the respawn window to prevent enemies from stealing their vehicle. 

This PR adds a check to see if the killed character was the last passenger in said vehicle, and if so then to unlock the car.


johnoclockdk:
This doesn't take into account whether the player is dead or not; it just unlocks the car when one player is in it.

QuentinGruber (changes requested):
https://github.com/QuentinGruber/h1z1-server/pull/2295#issuecomment-2545092278

GumpyH1Z1:
For some reason I can't push to this PR... so I'll do it this way.

Here is my suggestion on fixing the bug:


    if (client.vehicle.mountedVehicle) {
      const vehicle = this._vehicles[client.vehicle.mountedVehicle],
        container = vehicle?.getContainer();
    
      if (vehicle && container) {
        container.items = {
          ...container.items,
          ...client.character.getDeathItems(this),
        };
      }
    
      // Check if passenger is dead and is in driver seat
      const seatId = vehicle.getCharacterSeat(client.character.characterId);
      if (!client.character.isAlive && seatId === 0) {
        // Unlock car if driver is dead
        vehicle.setLockState(this, client, false);
        // Destroy the driver character, so it's possible to go into driver seat.
        client.character.destroy(this);
      }
    }
    
You could even make it more fancy and to be honest make it like it should, add a small timeout of about 5 seconds before the car unlocks and the driver disappears.



    if (client.vehicle.mountedVehicle) {
      const vehicle = this._vehicles[client.vehicle.mountedVehicle],
        container = vehicle?.getContainer();
    
      if (vehicle && container) {
        container.items = {
          ...container.items,
          ...client.character.getDeathItems(this),
        };
      }
    
      // Check if passenger is dead and is in driver seat
      const seatId = vehicle.getCharacterSeat(client.character.characterId);
      if (!client.character.isAlive && seatId === 0) {
        setTimeout(() => {
          // Unlock car if driver is dead
          vehicle.setLockState(this, client, false);
          // Destroy the driver character, so it's possible to go into driver seat.
          client.character.destroy(this);
        }, 5000);
      }
    }

With this, the bug should be fixed. I haven't tested this, so please do. I also made this piece of code assuming that it is not possible to lock the car from passenger seats 2 to 5, only from driver seat.

Nikolassparrow:
> This doesn't take into account whether the player is dead or not; it just unlocks the car when one player is in it.
Added the check for !isAlive. getPassengerList().length will return a 1 if there was only 1 person remaining in the car when you killed them, not a 0. So no the car will NOT unlock "when one player is in it."

Nikolassparrow:
> For some reason I can't push to this PR... so I'll do it this way.
> 
> Here is my suggestion on fixing the bug:
> 
> ```
> if (client.vehicle.mountedVehicle) {
>   const vehicle = this._vehicles[client.vehicle.mountedVehicle],
>     container = vehicle?.getContainer();
> 
>   if (vehicle && container) {
>     container.items = {
>       ...container.items,
>       ...client.character.getDeathItems(this),
>     };
>   }
> 
>   // Check if passenger is dead and is in driver seat
>   const seatId = vehicle.getCharacterSeat(client.character.characterId);
>   if (!client.character.isAlive && seatId === 0) {
>     // Unlock car if driver is dead
>     vehicle.setLockState(this, client, false);
>     // Throw out the driver character, so it's possible to go into driver seat.
>     this.dismountVehicle(client);
>   }
> }
> ```
> 
> You could even make it more fancy and to be honest make it like it should and add a small timeout of about 5 seconds before the car unlocks and the driver disappears.
> 
> ```
> if (client.vehicle.mountedVehicle) {
>   const vehicle = this._vehicles[client.vehicle.mountedVehicle],
>     container = vehicle?.getContainer();
> 
>   if (vehicle && container) {
>     container.items = {
>       ...container.items,
>       ...client.character.getDeathItems(this),
>     };
>   }
> 
>   // Check if passenger is dead and is in driver seat
>   const seatId = vehicle.getCharacterSeat(client.character.characterId);
>   if (!client.character.isAlive && seatId === 0) {
>     setTimeout(() => {
>       // Unlock car if driver is dead
>       vehicle.setLockState(this, client, false);
>       // Throw out driver character, so it's possible to go into driver seat.
>       this.dismountVehicle(client);
>     }, 5000);
>   }
> }
> ```
> 
> With this, the bug should be fixed. I haven't tested this, so please do. I also made this piece of code assuming that it is not possible to lock the car from passenger seats 2 to 5, only from driver seat.
> 
> **EDIT: Tested and it worked. Had to change the code a bit (this.dismountVehicle instead of destroy).**

Why would you "make it fancy" by adding a 5 second delay 😂😂 this practically defeats the purpose by still allowing a window of time to abuse the bug. Pointless addition.

Don't use getCharacterSeat, people can still abuse this bug by just swapping seats. That's why there needs to be a check for the amount of remaining passengers as this bug can still be abused by just swapping to another seat.

Also calling dismountVehicle in killCharacter causes the killed player to not be able to respawn (very very bad). So don't add that either.

Nikolassparrow:
@QuentinGruber This PR is good now.

johnoclockdk:
> > This doesn't take into account whether the player is dead or not; it just unlocks the car when one player is in it.
> 
> Added the check for !isAlive. getPassengerList().length will return a 1 if there was only 1 person remaining in the car when you killed them, not a 0. So no the car will NOT unlock "when one player is in it."

This is correct, but the dead player fills the seat, so you need to dismount the vehicle, I think.

Nikolassparrow:
> > > This doesn't take into account whether the player is dead or not; it just unlocks the car when one player is in it.
> > 
> > 
> > Added the check for !isAlive. getPassengerList().length will return a 1 if there was only 1 person remaining in the car when you killed them, not a 0. So no the car will NOT unlock "when one player is in it."
> 
> This is correct, but the dead player fills the seat, so you need to dismount the vehicle, I think.

Ahh I see what you mean now. I misunderstood what you were referencing, thank you. I've updated this PR with those changes.

PR #2297, opened 2024-12-16
Revert "Bugfix/locked vehicle on death"

QuentinGruber:
Reverts QuentinGruber/h1z1-server#2295 ffs @Nikolassparrow test your prs properly when the player respawn it respawn in the car

PR #2298, opened 2024-12-16
Fix bug - Locked vehicle on death & perma locked vehicle

GumpyH1Z1:
Fixed the issue where a vehicle would stay locked, when a player, which had locked the vehicle, would die and not respawn.

Fixes #2296

PR #2300, opened 2024-12-17
Fix for permanently locked vehicles (final)

Nikolassparrow:
#2295 and #2297 

Fixed the issue in #2297.

I found out you can just call dismountVehicle after sendKillFeed and it will do the same thing as I was trying to in #2295.


Nikolassparrow:
https://streamable.com/8067lv Video of it working 👍🏼

GumpyH1Z1:
This code isn’t correct yet. Right now it would unlock the vehicle, even if there are multiple occupants in the vehicle.

GumpyH1Z1:
> > This code isn’t correct yet. Right now it would unlock the vehicle, even if there are multiple occupants in the vehicle.
> 
> Bro what 😂😂 that's not even possible as isLocked = false doesn't get called unless !seatId, which only happens when the car is empty lmao
> 
> image

https://www.youtube.com/watch?v=foRmONOsqdY

Coming back on that picture, with the code shown, as dismountVehicle unlocks the vehicle by standard, but when someone dies its a different scenario as the vehicle should stay locked if there are other occupants in the vehicle and the vehicle was locked in the first place.

Hope that I informed you enough now. Also uploaded a video, showing my code is also working, don't know why you couldn't get it working...

QuentinGruber:
The changes in https://github.com/QuentinGruber/h1z1-server/pull/2298 will make it easier to maintain in the future.

QuentinGruber:
even tho you're way is easier

PR #2306, opened 2024-12-18
Fix for red tool cabinet respawning twice

Nikolassparrow:
Leftover code from an old pull request (#2045), got closed for some reason, but still is a bug in the game.


avcio9:
the duplicated models got removed from lootablePool a while ago

PR #2322, opened 2024-12-22
Removed vehicle locked when driver would get killed, but passengers a…

GumpyH1Z1:
As requested by Ultrax and Nick, I removed the ability of the vehicle to stay locked, when the driver was killed, but a passenger would be in the vehicle.


QuentinGruber:
i will implement that in https://github.com/QuentinGruber/h1z1-server/pull/2323

PR #2324, opened 2024-12-22
changeAlphaCrate

QuentinGruber:
- **change logic for "rare" reward**
- **first test of a changed alphaCrate**


zxywandxb111:
I think this affects the H1Z1 clone, it would be more reasonable to add a new H1EMU crate for rewards, although this increase the workload.

PR #2343, opened 2025-01-01
Add more ways to get reward crates

Nikolassparrow:
https://discord.com/channels/707525351357677610/1323668551407763508 - Also talked about pretty frequently that there should be more ways to get crates.

MaszuKa mentions breaking wooden crates, farming trees, and with a crowbar - also added when you kill someone in pvp.

Wasn't sure what to put the percentage at so I just put them all at 0.5% as that's the current chance for crates in lootable props.

Also was thinking that hitting a vehicle at all should give you a chance for a crate, but perhaps it'd be better to only allow it when a player gets scrap metal to prevent abuse?


zxywandxb111:
这个非常好，建议鼓励建造玩家  通常这种玩家不会去打架  只会建造         建议：砍树敲废车低概率获得  或者参考 JS游戏末期 搜集 三种颜色的拖鞋去合成     低概率刷新    鼓励搜物资 鼓励砍树敲铁的辛勤玩家

zxywandxb111:
This is very good, it's suggested to encourage players who focus on building. These players usually don't engage in fights and only focus on building Suggestion: Allow players to get resources with a low probability from chopping trees or destroying vehicles, or refer to the late stage of JS game where players collect different colored slippers to synthesize resources. Refresh with low probability, encourage players to search for resources, and motivate players who work hard at chopping trees and iron.

avcio9 (changes requested):
Items dropped by players have their spawnerId set to -1; you can use this instead of timeout to avoid abuse.
Most players tend to kill each other as soon as they meet, but we shouldn’t encourage this behavior among the few who don’t. The reward for killing/damaging should be removed.
The chance to receive a crate should only be calculated when you obtain metal from a car.
The chance to receive a crate when hitting a tree is too high; I think setting it to 1 is enough.
Now that we have other methods of obtaining crates, we can remove the chance during prop searches to prevent players from logging out and back in to reset the timer.

Nikolassparrow:
> Items dropped by players have their spawnerId set to -1; you can use this instead of timeout to avoid abuse. Most players tend to kill each other as soon as they meet, but we shouldn’t encourage this behavior among the few who don’t. The reward for killing/damaging should be removed. The chance to receive a crate should only be calculated when you obtain metal from a car. The chance to receive a crate when hitting a tree is too high; I think setting it to 1 is enough. Now that we have other methods of obtaining crates, we can remove the chance during prop searches to prevent players from logging out and back in to reset the timer.

So you want to remove this line of code from lootableProp?
Screenshot 2025-01-04 104857

avcio9 (approved):

PR #2344, opened 2025-01-01
Get suggestions for crates

zxywandxb111:
Recreate an official event from the late JS period where players could obtain crates. The refresh locations for the sandals and the crates received need to beed.


QuentinGruber:
obv this pr won't work @avcio9 but the idea of crafting a crate seems a good one to me, at least better than randomly looting containers

PR #2352, opened 2025-01-04
Fix: worn letters 

johnoclockdk:
I don't know if chance(num) is used somewhere else, but it was set too high, causing the worn letter to have a 0.02 change as given.


avcio9:
Hey, this is used for spawning crates, they have smaller chance than 1% so the function has to stay that way

johnoclockdk:
> Hey, this is used for spawning crates, they have smaller chance than 1% so the function has to stay that way

hey avico could you provide me with file

avcio9 (approved):

PR #2363, opened 2025-01-09
Fix PvE killing players with vehicle 

Nikolassparrow:
Reported by: purple in the discord

Currently on PvE, players are using vehicles to run players over and kill them. This PR fixes that, although I'm not a big fan of the solution, but it does work. 

The issue I was coming across was still applying fall damage for PvE players cause simply doing if (server.isPvE) return; won't give the user any fall damage. This was the only way I figured to work, but there's 110% a better way of doing this


QuentinGruber (approved):
at least it should work, seems good to me

PR #2369, opened 2025-01-10
Fix bushes being able to give crates

Nikolassparrow:
Slight oversight from #2343 

People found out pretty quickly that blackberry and regular bushes are pretty op for the new reward crate system.
This PR removes the ability for this to become the new meta, instead of its intended use as being a reward for meleeing trees.

### DIFF
--- a/src/servers/ZoneServer2016/managers/speedtreemanager.ts
+++ b/src/servers/ZoneServer2016/managers/speedtreemanager.ts
@@ -78,6 +78,7 @@ export class SpeedTreeManager {
     let count = 1;
     if (speedtreeDestroyed) return;
     let itemDefId = 0;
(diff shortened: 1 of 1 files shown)


johnoclockdk:
I do think that let could cause issues in the long run. why not just move `server.lootCrateWithChance(client, 1);` into the spawn log logic ?

Nikolassparrow:
> I do think that let could cause issues in the long run. why not just move `server.lootCrateWithChance(client, 1);` into the spawn log logic ?

And how exactly would this be bad in the long run😂?

QuentinGruber:
Hey @Nikolassparrow this pr will just await sometime, cz maybe we will rework this system but so i keep it open

PR #2373, opened 2025-01-12
Add debuff to spam crouching (enas)

Nikolassparrow:
Reported by: ysldrag in the discord (also very commonly talked about among H1EMU players)

Currently the meta in PvP (and even back in 2016) is to spam crouch, more commonly known as "enasing", to make your character very difficult to hit. I've talked to some other players and seen some suggested solutions to fix enasing, but more often involved a deeper understanding of client side modifying that I'm not versed in by any means.. 

I felt this was the best, quick solution to fix this major problem. I used existing code from how frostbite affects players to apply a slowness affect if a person reaches a crouchCount >= 5 within a certain time-frame. It lasts five seconds currently


Ultrax1995:
We tested with players on DeadZone. The best option is

duration: Date.now() + 1500,

johnoclockdk:
Nice work! There is a bug where crouching near a prop slows you down when you stand up.

QuentinGruber:
> Nice work! There is a bug where crouching near a prop slows you down when you stand up.
> 
> https://streamable.com/9zwfmy

still need to fix this, i guess it's bc it doesn't count the number of crouch but the time you're in crouch

QuentinGruber (approved):

PR #2378, opened 2025-01-15
Adrenaline shot

Nikolassparrow:
https://www.youtube.com/watch?v=jCbPwocMpx8 ~2:25 Got my information about adrenaline shot from this video.

Was reading some old h1z1 reddit posts and saw someone mention adrenaline shots, which H1EMU doesn't have currently. It's basically swizzle, but it can be found in the world (hospital), whilst swizzle you have to craft. 

Some of this code is a bit sloppy, particularly how I employed adrenaline after effects, but it does work so far and haven't found any bugs yet.


QuentinGruber:
using multiple ones in the same time allow to have infinite stamina for way too long is it normal ? also maybe we could add that the player get an heart attack when using multiple shot at the same time ? @Nikolassparrow

Nikolassparrow:
> using multiple ones in the same time allow to have infinite stamina for way too long is it normal ? also maybe we could add that the player get an heart attack when using multiple shot at the same time ? @Nikolassparrow

No, that's unintentional. I like the idea of adding a heart attack for having multiple active at the same time. How does adding a limit of three sound?

QuentinGruber:
> > using multiple ones in the same time allow to have infinite stamina for way too long is it normal ? also maybe we could add that the player get an heart attack when using multiple shot at the same time ? @Nikolassparrow
> 
> No, that's unintentional. I like the idea of adding a heart attack for having multiple active at the same time. How does adding a limit of three sound?

We could make that if adrenaline effect is there when you take a shot: you die but you can use another one during the after effect phase ? so you can use how many you want but you have to respect a timing ?

PR #2381, opened 2025-01-15
New feature (multiple 'owners') for base permissions

### LINKED ISSUE #2377: Multiple deck/tamper/shack 'owners'
**Is your feature request related to a problem? Please describe.**
Right now the only person that can add/remove people of a deck/tamper/shack, is the one that placed it. When that person is not around (sometimes this is for multiple days), it is impossible to add/remove others.

**Describe the solution you'd like**
My solution for this would be:
- Opening gates and doors on chest permission;
- Giving extra ownership to the ones on demolition permissions (which lets them invite/remove others onto or from the deck/tamper/shack, just not the original owner).

**Additional context**
I've already created a suggestion on this in the H1EMU Discord.
https://discord.com/channels/707525351357677610/1322982570962718760

GumpyH1Z1:
Closes #2377


GumpyH1Z1:
**Nice to Have:**
Update the GUI for other clients who have the permissions menu open simultaneously when a player is added or edited.

Currently, I cannot get this working because the packet `NpcFoundationPermissionsManagerBase`.ShowPermissions is only sent when interacting with or opening the menu. If a client already has the menu open, there is no way to check if it’s open or not, which means I can’t update the GUI for that specific client when a change occurs.

As a result, using `sendDataToAllOthersWithSpawnedEntity `triggers every client (with the correct permissions) to open the menu, instead of simply updating those who already have the menu open.

QuentinGruber (approved):

PR #2394, opened 2025-01-17
Up vehicle collision damage

Nikolassparrow:
https://discord.com/channels/707525351357677610/1329656945824759841
Suggested by s1vve, and meme to an extent (mentioned that vehicles take "basically nothing" flipping down the entire radio tower)

https://www.youtube.com/watch?v=M3kfkH5bids - guy takes cop car down entire radio mountain and has ~75% left, so I attempted to match these values. From what I tested, damage * 4 seemed to get pretty close when going top to bottom of radio tower


QuentinGruber:
@Nikolassparrow update your /dev, CI tests will always failed if you don't get the latest version of them.

QuentinGruber (approved):

zxywandxb111:
The natural decay of the vehicle also seems to be 4 times the original.

Nikolassparrow:
> The natural decay of the vehicle also seems to be 4 times the original.

That shouldn't be possible. Vehicle decay is managed separately from Vehicle Collision.

Issue #2380, opened 2025-01-15
Doors and gates not showing correct interactionstrings


GumpyH1Z1:
**Describe the bug**
In the current state of the game, doors and gates always display the message:
`[E] Open - Hold [E] Lock`.

However, this isn’t how it used to work, and to be honest, it’s quite annoying when you're trying to check if you’ve added codes to all gates and doors.

Previously:
- If you didn’t have permissions, it would display the message: `Enter access code` (adding [E] in front could be a nice touch).
- If you didn’t have permissions to lock gates and doors but had entered a code, it would show: `[E] Open`.

**Expected behavior**
- When looking at a gate/door **without permissions** to access, display the interaction string: `[E] Enter access code`.
- When looking at a gate/door **with a code applied or with appropriate permissions**, display: `[E] Open`.
- When looking at a gate/door **with permissions**, additional options like locking should behave as intended, display : `[E] Open - Hold [E] Lock`.

PR #2398, opened 2025-01-18
Fix gate/door interaction string

Nikolassparrow:
Fix for #2380 

Gumpy mentions that interaction strings for gates/doors should be based upon the appropriate permissions, so this PR fixes that.

The appropriate string for each type:
Owner of base / demolish permission: [E] Open - Hold [E] Lock
No granted access to door / no demo perms: Enter access code (having the interaction key at the beginning doesn't exist)
Granted access with no demo perms: [E] Open (target)


QuentinGruber:
Plz @Nikolassparrow update your branch with latest /dev I won't do it everytime

avcio9 (approved):

PR #2401, opened 2025-01-18
Randomize default skins

Nikolassparrow:
https://www.youtube.com/watch?v=vpChC-4cKhE 23:05 | 26:00 
https://www.youtube.com/watch?v=aFpSNGow3QI 51:00

So getRandomSkin() got removed in a previous PR, but it can still be utilized for randomizing default skins that should spawn in the world. Things like motorcycle helmets, 1000 bulk backpacks, and conveys used to spawn with different default skins such as: red/white conveys, orange/blue backpacks, or gray/white helmets.


src/servers/ZoneServer2016/managers/worldobjectmanager.ts
```diff
@@ -55,6 +63,30 @@ import { Npc } from "../entities/npc";
 import { scheduler } from "node:timers/promises";
 const debug = require("debug")("ZoneServer");
 
+export function getRandomSkin(itemDefinitionId: number) {
+  const allowedItems = [
+    Items.CONVEYS_BLUE,
+    Items.BACKPACK_BLUE_ORANGE,
+    Items.HELMET_MOTORCYCLE
+  ];
+  if (!allowedItems.includes(itemDefinitionId)) return itemDefinitionId;
```
QuentinGruber, line 72: would be easier to read if we just add a default case that return the original itemDef

Nikolassparrow: True 👍🏼

avcio9 (approved):

PR #2404, opened 2025-01-20
Feat: Clantags

johnoclockdk:
adding clantags

commands:
/clan create <tag>
/clan join <tag>
/clan accept <name>
/clan leave - leave current clan
/clan disband - only creator of the clan can do this


johnoclockdk:
any feedback would be much appreciated

GumpyH1Z1:
I have a suggestion that might be beyond the current scope, but could be fun to explore in the future. It would be great to have a GUI for managing invitations and deletions of other players. While adding a new tab in the inventory labeled 'clan' instead of 'group' might not be possible, an alternative approach could be to treat the group as a type of Foundation.

By doing so, you could leverage the existing UI for managing permissions in a Foundation. This could include using the buttons for build/demolish/storage/visitor rights to assign players the ability to invite or remove others from the clan (Foundation).

Additionally, it would be helpful if notifications, such as invitations to groups, were displayed at the top of the screen, allowing players to accept by pressing 'Y,' similar to the current group invitation system.

zxywandxb111:
After players join the team, add a green team name in front of their name that is visible to other players. Is it possible to achieve this in 2016 version?

johnoclockdk:
> After players join the team, add a green team name in front of their name that is visible to other players. Is it possible to achieve this in 2016 version?

if i color the nameplate then sure but idk yet

johnoclockdk:
@QuentinGruber i think i will close this one as i dont really feel like it should be added anymore

PR #2412, opened 2025-01-22
Fix orange backpack bulk & not skinnable issue

Nikolassparrow:
image
Reported by: Wxlmart in the discord

Orange backpack model 1355 has several bugs to it, never use it (prob some old shit dbg was too lazy to remove) 

Changed the model to 2116 which has no bugs, current bulk, and can be skinned.


QuentinGruber (approved):
thx

PR #2425, opened 2025-01-26
Fix metric survive time

Nikolassparrow:
Currently startedSurvivingTP will display times during when the player was logged off, leading to bigger than intended times. This PR fixes that by "shifting" the time to prevent offline time being tracked in startedSurvivingTP


QuentinGruber:
are you sure lastLoginDate is kept updated ?

Nikolassparrow:
> are you sure lastLoginDate is kept updated ?

From testing and what I've seen in the code, yes it looks maintained

QuentinGruber:
i tested it with mongo and somehow it wasn't saving time... Will retry another day

QuentinGruber (approved):

PR #2427, opened 2025-01-27
Add water bottle filling while raining in dew collectors

Nikolassparrow:
Noticed that dew collectors don't detect when rain is currently on to give a multiplier for dew collectors, so this PR adds that. Seems to work good so far from testing

### DIFF
--- a/src/servers/ZoneServer2016/managers/smeltingmanager.ts
+++ b/src/servers/ZoneServer2016/managers/smeltingmanager.ts
@@ -284,8 +284,9 @@ export class SmeltingManager {
   ) {
     const currentDate = new Date(server.inGameTimeManager.time * 1000);
     const currentHour = currentDate.getUTCHours();
-    const isRaining = server.weatherManager.rainingHours.includes(currentHour) || 
-      server.weatherManager.globalPrecipation > 1 || 
+    const isRaining =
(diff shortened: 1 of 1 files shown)


QuentinGruber:
Idk how rainingHours works so i let this to @avcio9

src/servers/ZoneServer2016/managers/smeltingmanager.ts
```diff
@@ -282,6 +282,14 @@ export class SmeltingManager {
     subEntity: CollectingEntity,
     container: LoadoutContainer
   ) {
+    const currentDate = new Date(server.inGameTimeManager.time * 1000);
+    const currentHour = currentDate.getUTCHours();
+    const isRaining =
+      server.weatherManager.rainingHours.includes(currentHour) ||
+      server.weatherManager.globalPrecipation > 1 ||
```
QuentinGruber, line 289: why not just using .rain ? and also this way it's inconsistent it could rain for 4 mins and if the rain stop and the smeltablemanager run right after it won't ack that it was raining

QuentinGruber (changes requested):

QuentinGruber:
https://github.com/QuentinGruber/h1z1-server/pull/2458

PR #2434, opened 2025-01-31
Add durability damage to tools when meleeing

Nikolassparrow:
https://h1z1stuff.weebly.com/melee-weapons.html
Got all my information about durability, and their damages here ^


QuentinGruber (changes requested):
Seems bad to have all this code duplicated, there must be another way of doing that

Nikolassparrow:
> Seems bad to have all this code duplicated, there must be another way of doing that

Yea I agree. Also is just ugly to have it duplicated like that

QuentinGruber (approved):
About duplication, OnMeleeHit is inherited from the baseentity class so you've could add that to that class and call super() in each OnMeleeHit to avoid duplication. But honestly it's good for me in this current shape @Nikolassparrow

PR #2436, opened 2025-01-31
Add durability loss while walking in shoes

Nikolassparrow:
So currently shoes have durability, but no use for them. This PR adds durability loss to shoes when walking. 

Ideally, the damage given to the shoe should be based upon the materialType that the player is currently walking on, but this is good for now. 

From testing these were values that felt "fair" but could be adjusted of course.

Boots: 1.2 dmg per step - 12,000 steps 
Conveys : 0.4 dmg per step - 13,500 steps

### DIFF
--- a/src/servers/ZoneServer2016/zonepackethandlers.ts
+++ b/src/servers/ZoneServer2016/zonepackethandlers.ts
@@ -1665,6 +1665,9 @@ export class ZonePacketHandlers {
       // Check current interaction
       client.character.checkCurrentInteractionGuid();
 
+      // Handle shoe durability
+      client.character.handleShoeDurability(server);
+      
       // Timeout interaction lock UI
       if (
         client.character.lastInteractionRequestGuid &&
--- a/src/servers/ZoneServer2016/zonepackethandlers.ts
+++ b/src/servers/ZoneServer2016/zonepackethandlers.ts
@@ -1667,7 +1667,7 @@ export class ZonePacketHandlers {
 
       // Handle shoe durability
       client.character.handleShoeDurability(server);
-      
+
       // Timeout interaction lock UI
       if (
         client.character.lastInteractionRequestGuid &&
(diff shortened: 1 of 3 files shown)


src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -1665,6 +1665,9 @@ export class ZonePacketHandlers {
       // Check current interaction
       client.character.checkCurrentInteractionGuid();
 
+      // Handle shoe durability
+      client.character.handleShoeDurability(server);
```
QuentinGruber, line 1669: Sadly calling this function here use too much cpu, it took up to 0.200 ms  while the whole processing of the packet without this func take 0.02-0.04ms. This is the most used packet in the game so that may be bad to add that

QuentinGruber: Also it means if a player run in circle instead of a straight line ( who send less positions updates ) it will damage his boots faster

QuentinGruber (changes requested):
I like the idea but idk what to think about this implementation. maybe we can found another way of doing it

johnoclockdk:
i do also like the idea but 13,500 seem abit to low for me but havent tested it will do later

src/servers/ZoneServer2016/zonepackethandlers.ts
Nikolassparrow: I see. Yea that's really bad especially if a lot of other players would be triggering this packet a bunch too. 

Maybe a method where we "collect" damage over a certain amount of time (5 seconds perhaps) and then apply the damage to the shoe? Or would this still cause the high cpu usage?

QuentinGruber: > I see. Yea that's really bad especially if a lot of other players would be triggering this packet a bunch too.
> 
> Maybe a method where we "collect" damage over a certain amount of time (5 seconds perhaps) and then apply the damage to the shoe? Or would this still cause the high cpu usage?

that would still be better but on which factor will we apply the damage ?

QuentinGruber:
we would have to find another way of doing this. Maybe by doing a routine every x secs and if the player has sprinted or something during that time frame we apply damages... it's not ideal but it won't cost us most in cpu.

PR #2438, opened 2025-02-02
Add radio functionality (WIP)

Nikolassparrow:
So I was messing around in the RadioLoadout actionscript file, I was aiming to get the RadioChannel packet working with success. I'm not versed in reversing packets, so there could be other fields, but radioType and currentChannel are the two I got working. 

radioType might be the voice channel type (proximity, group, radio etc), when I turned on the radio, radioType = 6
currentChannel handles the channel the player currently has selected from the radio channel slider.

I have a rough idea how to get the radio voice chat to work, but I'm not confident in how voice chat works in H1EMU, so perhaps someone could take a look into getting that to work :D


avcio9:
radio is working, it reads the values from games memory

Issue #1665, opened 2023-11-02
World placed objects overhaul


Z1Meme:
- All freeplace entities in the world should be easily destructable to prevent griefing (not too many melee or projectile hits)

- Freeplace entities in the world should have heavier decay

- Freeplace entities should have a "protectionParent" assigned if they are not directly parented to another base part, meaning that they'll disappear if that parent respawns, and they are protected from extra damage / decay unless a player has demo perms on the parent.

- foundations / tampers / expansions need to check for lootableConstruction under placement area on placement

- Hidden player / container security in shelters should apply to all entities in a shelter (would also block explosions outside from killing players inside)

- All placable entities and characters should have a "securedShelter" variable that is the characterId of the shelter containing them, in order to check if they should / should not spawn for other players. Change "isHidden" back to a boolean and use that only for vanished characters / spectating.

PR #2443, opened 2025-02-06
Hide freeplaced entities inside secured ConstructionChildEntities 

Nikolassparrow:
https://streamable.com/4245k3 (sorry about music) 

Currently freeplaced entities are visible to players outside of secured shelters/shacks. Players are hidden which is great, but most, if not all entities inside of a secured shelter/shack should be hidden. Work is still needed on #1665 for adding a "securedShelter" variable, but using existing isHidden architecture, can still accomplish our goals.

Ideally this should be tested on DeadZone, or another private server to insure there are no strange bugs or exploits tied in. From testing, it felt stable minus a small drawback in time - that should be looked into so it doesn't cause unnecessary lag on the server.


QuentinGruber:
from what i saw it doesn't seems to affect performances. we will test that during the next week

src/servers/ZoneServer2016/managers/constructionmanager.ts
```diff
@@ -2008,6 +2024,57 @@ export class ConstructionManager {
...
+          let hasVisitPermission = false;
+          if (constructionEntity) {
+            hasVisitPermission = constructionEntity.getHasPermission(
+              server,
+              iteratedClient.character.characterId,
+              ConstructionPermissionIds.VISIT
+            );
+          }
+
+          const isSameGroup =
```
QuentinGruber, line 2052: why does it has to be in the same group ?

QuentinGruber (changes requested):

src/servers/ZoneServer2016/managers/constructionmanager.ts
Nikolassparrow: It's also in constructionHidePlayer. From testing on my own, it's so players can still see their teammates if they're inside a specific shelter/part of base. This would be useful if you lived in a big base and your teammate asks: "hey come to me" you can know where is at.

johnoclockdk: > It's also in constructionHidePlayer. From testing on my own, it's so players can still see their teammates if they're inside a specific shelter/part of base. This would be useful if you lived in a big base and your teammate asks: "hey come to me" you can know where is at.

@QuentinGruber maybe give the man an reply ppl want to have this ingame

https://discord.com/channels/707525351357677610/1361406050099662958

QuentinGruber: > > It's also in constructionHidePlayer. From testing on my own, it's so players can still see their teammates if they're inside a specific shelter/part of base. This would be useful if you lived in a big base and your teammate asks: "hey come to me" you can know where is at.
> 
> @QuentinGruber maybe give the man an reply ppl want to have this ingame
> 
> https://discord.com/channels/707525351357677610/1361406050099662958

i just need to test this before rolling it out

nymanwow: Hows it going on this one, can we push it to test server?

RumTumEric:
Working pretty good so far :)

QuentinGruber:
> Working pretty good so far for Storage Containers and Furnaces. Although both Workbench and Weapon Workbench bars are still visible from the outside. This applies to both shacks and shelters with protection.

it's good enough or we need to fix that before merging this ?

RumTumEric:
It's better but, if possible, I think working a fix on all placed entities would be worth the wait.

QuentinGruber (approved):

PR #2569, opened 2025-04-22
Added skinned guns to melt table

ZonnyLee:
* Also added the R380 to the melt table
* Added M1911 to Dismantle next to weapon workbench for gun parts
   Not able to add multiple guns to dismantle into gun parts due to errors


ZonnyLee:
Tested melting all skins, R380 and the M1911 to gun parts

PR #2577, opened 2025-04-25
Nerf mollys

johnoclockdk:
Nerf mollys
and added .env to .gitignore


johnoclockdk:
i did also try to apply armor reduction didnt work like i want it

PR #2613, opened 2025-05-29
Refresh players inventory after exchange

ZonnyLee:
Should ensure that player inventory gets refreshed after an exchange

Needs testing to make sure it doesent brake something @QuentinGruber


ZonnyLee:
@QuentinGruber I have not tested it myself

QuentinGruber:
> @QuentinGruber I have not tested it myself

i did and i saw nothing bad but idk if it solved anything ( and i believe we already try that a long time ago )

PR #2616, opened 2025-05-31
Crate alert fix

nymanwow:
Added enums for new crates

Edit h1emuex enum to h1emuexclusive

Added function to correct the crate name using giverewardtoall


nymanwow:
Will fix prettier when I'm at pc

PR #2617, opened 2025-06-01
/giverewardtoall adjustment

nymanwow:
Added the possibility to reward multiple crates at once with /giverewardtoall command

Changed the alert when doing the command


nymanwow:
Closed, no needed change.

PR #2618, opened 2025-06-02
Adjusted giverewardtoall command, possible to add more than one crate ID in one command 

nymanwow:
Adjusted giverewardtoall command, possible to add more than one crate ID in one command


RumTumEric:
After discussing with fellow mods and devs, would it be possible to implement a limit to prevent accidently fat fingering 5000x crates to every player on the server?

nymanwow:
> After discussing with fellow mods and devs, would it be possible to implement a limit to prevent accidently fat fingering 5000x crates to every player on the server?

Yes m8, i did remove the count option after your suggestion earlier :D

QuentinGruber (dismissed):

QuentinGruber (dismissed):

QuentinGruber (approved):

PR #2619, opened 2025-06-04
Disable emote

johnoclockdk:


src/servers/ZoneServer2016/handlers/commands/commands.ts
```diff
@@ -443,16 +443,18 @@ export const commands: Array<Command> = [
...
-      server.sendDataToAllWithSpawnedEntity(
-        server._characters,
-        client.character.characterId,
-        "Animation.Play",
-        {
-          characterId: client.character.characterId,
-          animationId: animationId
-        }
-      );
+      if (server.isPvE) {
```
QuentinGruber, line 446: disable the whole function not just the end

QuentinGruber (changes requested):

PR #2622, opened 2025-06-08
Adjustment + new (FOR NEXT WIPE)

nymanwow:
Removed the ability to build close to warehouse in Pleasent Valley

Added an POI on the I2 camp (NE Ranchito)

Extended the POI of H2 camp (NW Ranchito)

Removed the ability to build under bridges since this issue occured during the wipe and people were sent out of their vehicles because of permission.


nymanwow:
@QuentinGruber 

This would be added til next wipe, since there is some bases already built at the new poi's ,

QuentinGruber:
> @QuentinGruber
> 
> This would be added til next wipe, since there is some bases already built at the new poi's ,

yep ofc :stuck_out_tongue:

PR #2650, opened 2025-06-28
allow some default items to be in lootbag

QuentinGruber:
### TL;DR

Improved loot bag generation by using a whitelist approach for valid items.

### What changed?

- Created a new array `invalidItemForLootbag` that explicitly lists items that should not be included in loot bags
- Added a new function `isValidForLootbag()` to check if an item should be included in a loot bag
- Replaced calls to `isDefaultItem()` with the new `isValidForLootbag()` function when determining what items to include in loot bags

### How to test?

1. Kill a character that has items in their inventory
2. Verify that a loot bag is created
3. Check that the loot bag does not contain any of the invalid items (fists, flashlight, map, improvised compass, gray/blue boots)
4. Verify that all other valid items are properly included in the loot bag

### Why make this change?

This change provides a more explicit and maintainable approach to determining which items should be included in loot bags. By using a specific list of invalid items rather than a generic "default item" check, we can more precisely control what items players can loot, improving gameplay balance and reducing potential exploits.

### DIFF
--- a/src/servers/ZoneServer2016/entities/basefullcharacter.ts
+++ b/src/servers/ZoneServer2016/entities/basefullcharacter.ts
@@ -63,6 +63,24 @@ function getGender(actorModelId: number): number {
   }
 }
 
+const invalidItemForLootbag: Items[] = [
+  Items.WEAPON_FISTS,
+  Items.WEAPON_FLASHLIGHT,
+  Items.MAP,
+  Items.COMPASS_IMPROVISED,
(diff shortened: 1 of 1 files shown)


QuentinGruber:
### Merge activity

* **Jun 29, 4:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/QuentinGruber/h1z1-server/2650).

PR #2676, opened 2025-07-06
Reworked getDeathItems logic

jasonmoonen:
Introduced PlayerDeathItemConfig to enable configuration of item IDs and item classes for applying damage to items when a player dies. If an item's durability drops to 0, the configured lootItems are added to the lootbag. Also resolved an issue that prevented items without durability from being included in the lootbag.


johnoclockdk:
this is awsome

QuentinGruber (approved):

PR #2686, opened 2025-08-15
Fix grinder desync issue

jasonmoonen:
It looks like sending Items.SetAccountItemManager removes the ghost items.

During testing, I found that UpdateAccountItem executes without any issues, but RemoveAccountItem does not always run on the client.


jasonmoonen:
@QuentinGruber

QuentinGruber (approved):

PR #2706, opened 2025-09-30
Reduce plane movement speed from 80000 to 40000

johnoclockdk:


jasonmoonen (changes requested):
Did you test this? If my memory serves me right the higher the value, the slower it is. So this makes the plane go faster

JayyWayz:
Isn't this in the config?
I think it's set to 140000 there

johnoclockdk:
> Did you test this? If my memory serves me right the higher the value, the slower it is. So this makes the plane go faster

i will test this later to check

johnoclockdk:
> Isn't this in the config? I think it's set to 140000 there

no config but ill add it

johnoclockdk:
> > > Isn't this in the config? I think it's set to 140000 there
> > 
> > 
> > no config but ill add it
> 
> Does this not work yet? image_2025-09-30_162435792

you right

PR #2732, opened 2025-12-10
Refactor emote command to use emote names

johnoclockdk:
The /emote command now accepts emote names instead of IDs and checks ownership for restricted emotes. Added /emotelist command to display available emotes. This improves usability and enforces emote access based on account items.


johnoclockdk:
@QuentinGruber I didn’t make a JSON file but a .ts file instead. I hope that’s okay. if you want a json let me know and ill change it

PR #2751, opened 2025-12-21
this should fix the voice chat for private servers

johnoclockdk:


src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -859,7 +859,7 @@ export class ZonePacketHandlers {
         if (!client.characterReleased) return;
         if (client.firstReleased) {
           server.sendData<H1emuVoiceInit>(client, "H1emu.VoiceInit", {
-            args: `${server._loginServerInfo.address} ${server._worldId}`
+            args: `${server.voiceChatManager.serverAddress} ${server._worldId}`
```
QuentinGruber: also it should be unset by default not to 127.0.0.1 and when unset default to our voice server. Kinda how the loginserver configuration for a zoneserver work. This way community servers won't loose access to our voice server

PR #2775, opened 2026-02-05
Remove unnecessary logging, fix interaction glow, and ensure doors break when shelters are destroyed

jasonmoonen:


src/servers/SoeServer/soeserver.ts
```diff
@@ -451,7 +451,7 @@ export class SOEServer extends EventEmitter {
         }
       }
     } catch (e) {
-      console.log(e);
+      debug(e);
```
QuentinGruber: This one should keep being a console log since the process exit after it

src/servers/SoeServer/soeserver.ts
```diff
@@ -97,8 +97,8 @@ export class SOEServer extends EventEmitter {
         performance.measure("A to B", "A", "B");
       }, intervalTime);
     } catch (e) {
-      console.error(e);
-      console.log("PerformanceObserver not available");
+      debug(e);
+      debug("PerformanceObserver not available");
```
QuentinGruber: Should be trigger once anyway it could be kept as a console.log

QuentinGruber (approved):

PR #2777, opened 2026-02-08
Bridge POI's

illudedPerception:
Bridge POI's to prevent people building under bridges and people getting kicked from cars

Added POI's to
Misty peak dam (all 3 bridges)
Cranberry bridge
Eastwood stop bridge
Rancho bridge w/ Magpie Rd bridge

Moved an addon POI from kurama to military (under the right category)


data/2016/zoneData/Z1_POIs.json
```diff
@@ -428,12 +469,6 @@
         [ 1966.49, -2661.38 ],
         [ 1600.39, -2674.76 ],
         [ 1615.91, -2992.76 ]
-      ],
```
illudedPerception, line 431: this was moved to where it should be (under military base)

jasonmoonen (commented):
Maybe we should go over the poi blocks completely while we're at it

illudedPerception:
> Maybe we should go over the poi blocks completely while we're at it

Thats the system what would allow closer building for pve and would still protect bridges ?

illudedPerception:
maybe just let this be a temp fix until you get to the real system or if you do, maybe the C server will be done before that

PR #2778, opened 2026-02-11
See mound on fertilized tillers on hover + small fixes

looc343:
New feature to help identify what tillers have been fertilized. When looking on a plant on a tiller that has been fertilized the small mound will appear for 3 minutes for the player.

Fixed a bug where planting a new plant on a fertilized tiller didn't reduce the first growth stage.

Slightly increased the range where you can pick up proximity items so that you always reach the ones displayed inside the inventory. 

When sending createcomponent repdata to non world items we dont need to send the entire buffer since the last byte is not used. Now we only send the required amount of bytes to include name ID. (This doesn't appear to break anything but we'll see)

### DIFF
--- a/src/packets/ClientProtocol/ClientProtocol_1080/replication.ts
+++ b/src/packets/ClientProtocol/ClientProtocol_1080/replication.ts
@@ -23,7 +23,15 @@ export function packComponentNameString(name: string) {
 }
 
 export function packClientNpcComponent(obj: any) {
-  const raw: Buffer = Buffer.alloc(78);
+  let raw: Buffer;
+  
+  if (obj["worldItem"]) {
+    raw = Buffer.alloc(78);
+    raw.writeUint8(1, raw.length - 1);
+  }
+  else {
+    raw = Buffer.alloc(16);
+  }
   if (obj["nameId"]) {
     raw.writeUInt32LE(obj["nameId"], 12);
   }
@@ -31,7 +39,6 @@ export function packClientNpcComponent(obj: any) {
   raw.writeUint8(122, 0);
   raw.writeUint8(2, 1);
 
-  raw.writeUint8(1, raw.length - 1);
   return raw;
 }
 
--- a/src/servers/ZoneServer2016/entities/plant.ts
+++ b/src/servers/ZoneServer2016/entities/plant.ts
@@ -32,6 +32,9 @@ export class Plant extends ItemObject {
 
   /** Next time (milliseconds) that the crop will enter the next state */
   nextStateTime: number;
+  
+  /** Next time (milliseconds) that a fertilizer mound can appear */
+  nextMoundTime: number = 0;
 
   /** Time (milliseconds) it takes for a crop to enter the next state - Default: 8hrs */
   readonly growTime = 28800000;
@@ -188,8 +191,9 @@ export class Plant extends ItemObject {
   }
 
   OnInteractionString(server: ZoneServer2016, client: ZoneClient2016): void {
-    if (this.isFertilized) {
+    if (this.isFertilized && new Date().getTime() > this.nextMoundTime) {
       const pos = this.state.position;
+      this.nextMoundTime = new Date().getTime() + 180000;
       server.sendData<CharacterPlayWorldCompositeEffect>(client, "Character.PlayWorldCompositeEffect",
         {
           characterId: this.characterId,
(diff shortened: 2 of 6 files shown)


src/packets/ClientProtocol/ClientProtocol_1080/replication.ts
```diff
@@ -23,15 +23,22 @@ export function packComponentNameString(name: string) {
 }
 
 export function packClientNpcComponent(obj: any) {
-  const raw: Buffer = Buffer.alloc(78);
+  let raw: Buffer;
+  
+  if (obj["worldItem"]) {
```
jasonmoonen, line 28: Does it only save data or does it actually have purpose?

src/servers/ZoneServer2016/entities/plant.ts
```diff
@@ -78,7 +81,10 @@ export class Plant extends ItemObject {
     const parent = server._temporaryObjects[
       parentObjectCharacterId
     ] as PlantingDiameter;
-    if (parent.isFertilized) this.isFertilized = true;
+    if (parent.isFertilized) {
+      this.isFertilized = true;
+      this.nextStateTime = new Date().getTime() + this.growTime / 2;
```
jasonmoonen, line 86: Why was this added here?

src/servers/ZoneServer2016/zonepackethandlers.ts
```diff
@@ -2786,7 +2786,7 @@ export class ZonePacketHandlers {
 
       if (
         !isPosInRadius(
-          server.proximityItemsDistance,
+          server.proximityItemsDistance + 0.2, // It doesn't always reach otherwise
```
jasonmoonen, line 2789: Why not increase proximityItemDistance instead?

jasonmoonen (changes requested):

src/packets/ClientProtocol/ClientProtocol_1080/replication.ts
looc343: what do you mean? for worlditems the last bool needs to be 1, for other items this doesnt matter so we dont need to send the last 60 bytes

src/servers/ZoneServer2016/entities/plant.ts
looc343: if the tiller is already fertilized when the seed is placed the plant would not know its fertilized and get the full 8 hours added to the timer before, now it adds only 4.

src/servers/ZoneServer2016/zonepackethandlers.ts
looc343: well both the range of what items are shown and the pickup range is proximityItemDistance. When they are the same you cant pick up the farthest away items for some reason, maybe rounding or something. So the range to pickup needs to be slightly longer for it to work reliably

src/packets/ClientProtocol/ClientProtocol_1080/replication.ts
jasonmoonen: Did you find the read function that works like this? From my research, the client doesn’t differentiate between these values, so they may still be needed. Other functions also accept the data even though the struct isn’t fully correct, which might also be happening here.

src/servers/ZoneServer2016/entities/plant.ts
jasonmoonen: Aha! I will test, good catch!

src/packets/ClientProtocol/ClientProtocol_1080/replication.ts
looc343: From my understanding FUN_140addc90 is responsible for deserializing the buffer into the component struct on the client. If the buffer doesn't contain all expected fields it will just copy a 0 instead. So sending the last 60 bytes as 0 or not sending them at all should populate the internal struct in the same way.

jasonmoonen: From the function you provided, it reads 78 bytes. Even if the client can handle whether it’s sent or not, it’s still better to implement the full struct properly. In that case, this 

> Buffer.alloc(78)

 should be replaced with the struct.

looc343: OK, your call

johnoclockdk:
bump

PR #2779, opened 2026-02-23
Prevent containers and workbench to be visible through walls + crafting QOL

looc343:
Tried to make some changes to prevent containers and workbenches to be seen through walls in some scenarios. This is to prevent raiders to use glitches and exploits to see what rooms might have loot. 

First I delayed spawning these items until later when gates and doors load to prevent seeing them when logging in.

Second change is to only show the health bar of these items for players with container permissions.

Lastly I reused code from checkBuildingObstruct to hide proximity container items when blocked by a wall.

Also changed so overflowing items from lootContainerItem will be put inside a mounted container instead of dropped. Now you can afk craft and the items will always stay inside the container and not be dropped.


QuentinGruber (dismissed):

BryanMacleod:
I removed my last comment as i just experienced the glitch. This is not the bug i was related too ;) 

Good work!

jasonmoonen (dismissed):

Z1Meme:
Too many merge conflicts on this PR, new PR --> https://github.com/QuentinGruber/h1z1-server/pull/2887

PR #2789, opened 2026-04-06
explosion effect fix

BryanMacleod:
- All explosives (IED, landmine, ethanol, biofuel) now go through a proper explosion queue
- Chain explosions process correctly without race conditions
- IED fuse effect shows reliably before detonation


johnoclockdk:
The main server is using the raytrace plugin, so maybe add a check to disable this behavior when it’s enabled. Also, why create a whole new system?

PR #2790, opened 2026-04-06
fill bottles loop + consumable auto-loop

BryanMacleod:
- Interacting with a water source auto-fills all empty bottles one by one, stops on move
- Using a bottle from inventory while standing in water auto-fills all bottles
- Using food/drink/comfort items now auto-loops until resources are full
- Canned food loops regardless of hunger or comfort (used for metal scrapping)
- Medical items auto-loop until full HP; shows correct message if already full or healing in progress

### DIFF
--- a/src/servers/ZoneServer2016/entities/watersource.ts
+++ b/src/servers/ZoneServer2016/entities/watersource.ts
@@ -119,12 +119,7 @@ export class WaterSource extends TaskProp {
     }
   }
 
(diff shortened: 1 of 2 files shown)


jasonmoonen:
Same applies here. The auto‑loop for medical items would give an unfair advantage, so I’m going to close it. It can still be included as an example for plugins though.

PR #2791, opened 2026-04-06
Add /closegates command with deck permission support and 20 door limit

BryanMacleod:


johnoclockdk:
i dont think this is good idea  this can be abuse

BryanMacleod:
> i dont think this is good idea  this can be abuse

Only need to lock the command on raid.

jasonmoonen:
Good idea but not vanilla. Would love to include this in the upcoming examples for plugins

PR #2799, opened 2026-05-18
0.47.4-10

jasonmoonen:


QuentinGruber (dismissed):

QuentinGruber:
* Maybe the anti ddos on weapon packets is to small, i've triggered it with a simple AR spray
* It won't conflict with the raycast plugin ? or we remove the raycast plugin ?

@jasonmoonen

jasonmoonen:
> * Maybe the anti ddos on weapon packets is to small, i've triggered it with a simple AR spray
> 
> * It won't conflict with the raycast plugin ? or we remove the raycast plugin ?
> 
> 
> 
> @jasonmoonen

It should only measure incoming packets so it shouldn't have an effect on raycast. I will do some more tests with weapons!

PR #2801, opened 2026-05-23
Fix landmines triggering IEDs

jasonmoonen:


QuentinGruber (approved):
gfm but it's not a wanted behavior ?

jasonmoonen:
Yes, it still does. But now when there's landmines on the map (not even near) that will turn IEDs into Landmines.

PR #2809, opened 2026-06-09
Add more apm spans and splits

jasonmoonen:


src/servers/ZoneServer2016/managers/worlddatamanager.ts
```diff
@@ -185,6 +212,11 @@ export class WorldDataManager {
     }
   }
 
+  private requireFresh(path: string) {
```
QuentinGruber: this could be a global util

src/servers/ZoneServer2016/managers/worlddatamanager.ts
```diff
@@ -235,16 +235,13 @@ export class WorldDataManager {
...
-      await Promise.all([
-        this.loadVehiclesData() as Promise<FullVehicleSaveData[]>,
-        this.loadConstructionData() as Promise<ConstructionParentSaveData[]>,
-        this.loadWorldFreeplaceConstruction() as Promise<
-          LootableConstructionSaveData[]
-        >,
-        this.loadCropData() as Promise<PlantingDiameterSaveData[]>,
-        this.loadTrapData() as Promise<TrapSaveData[]>
-      ]);
+    const vehicles = (await this.loadVehiclesData()) as FullVehicleSaveData[];
```
QuentinGruber: i get it that for saving we don't do it concurrently but why when fetching too ?

QuentinGruber (approved):

PR #2812, opened 2026-06-14
Support multi-variant crafting recipes

### LINKED ISSUE #2545: Gun parts
Currently you can only craft gun parts from an AR15, could you make it that all weapons can be broken down for parts.

https://discord.com/channels/707525351357677610/1320791950630060142

johnoclockdk:
Allow recipes to be either a single Recipe or an array of Recipe variants

### DIFF
--- a/src/servers/ZoneServer2016/entities/character.ts
+++ b/src/servers/ZoneServer2016/entities/character.ts
@@ -312,7 +312,7 @@ export class Character2016 extends BaseFullCharacter {
   /** HashMap of all recipes on a server
    * uses recipeId (number) for indexing
    */
-  recipes: { [recipeId: number]: Recipe | Recipe[] } = recipes;
+  recipes: { [recipeId: number]: Recipe } = recipes;
 
   currentChallenge: ChallengeType = ChallengeType.NONE;
 
@@ -434,8 +434,15 @@ export class Character2016 extends BaseFullCharacter {
     const recipes: Array<any> = [];
     let i = 0;
     for (const recipeEntry of Object.values(this.recipes)) {
-      const recipeDef = server.getItemDefinition(Number(recipeKeys[i]));
-      if (!recipeDef) continue;
+      const recipeId = Number(recipeKeys[i]);
+      const recipe = Array.isArray(recipeEntry) ? recipeEntry[0] : recipeEntry;
+
+      // Use rewardId if available, otherwise use the recipe key (item ID)
+      const itemId = recipe?.rewardId || recipeId;
+      const recipeDef = server.getItemDefinition(itemId);
+
+      if (!recipeDef) {
+      }
       if (
         server.isBattleRoyale() &&
         ![Items.BANDAGE, Items.BACKPACK_SATCHEL, Items.ARMOR_PLATED].includes(
@@ -456,11 +463,9 @@ export class Character2016 extends BaseFullCharacter {
         variantIndex++
       ) {
         const recipe = recipeArray[variantIndex];
-        // Use variant index in recipeId to differentiate multiple recipe variants
+        // Use recipe ID (not item ID) as base for variant index calculation
         const variantRecipeId =
-          recipeArray.length > 1
-            ? recipeDef.ID + variantIndex * 1000000
-            : recipeDef.ID;
+          recipeArray.length > 1 ? recipeId + variantIndex * 1000000 : recipeId;
 
         recipes.push({
           recipeId: variantRecipeId,
--- a/src/servers/ZoneServer2016/entities/character.ts
+++ b/src/servers/ZoneServer2016/entities/character.ts
@@ -429,20 +429,14 @@ export class Character2016 extends BaseFullCharacter {
   }
 
   pGetRecipes(server: ZoneServer2016): any[] {
-    const recipeKeys = Object.keys(this.recipes);
-
     const recipes: Array<any> = [];
-    let i = 0;
-    for (const recipeEntry of Object.values(this.recipes)) {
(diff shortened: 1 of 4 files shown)


johnoclockdk:
@jasonmoonen @QuentinGruber Probably best if you both approve this before merging, as it may introduce new dups.

src/servers/ZoneServer2016/managers/craftmanager.ts
```diff
@@ -102,6 +102,19 @@ function getCraftComponentsDataSource(
...
 
+/**
+ * Helper function to get a recipe, handling both single recipes and recipe arrays
+ * @param recipe A Recipe or Recipe[] from the recipes object
+ * @returns The first recipe if it's an array, or the single recipe, or undefined
+ */
+function getRecipeFromEntry(recipe: any): Recipe | undefined {
+  if (!recipe) return undefined;
+  if (Array.isArray(recipe)) {
+    return recipe[0]; // Return the first recipe from the array
```
jasonmoonen: Not sure about this to be honest

jasonmoonen (dismissed):

src/servers/ZoneServer2016/managers/craftmanager.ts
johnoclockdk: hmm any idea what you do then ?

jasonmoonen: Not really, what I do know is that Daybreak used incremental recipe ids (you should be able to find some ids that I found in enums.ts) instead of using the item ids for recipes, maybe if we start using it we might not need it to be an array. But I might be wrong and this is good enough. I'll let @QuentinGruber decide. Maybe he has a better idea

johnoclockdk: ill see if i cant get a better way later tonight but let see what @QuentinGruber says 👍

src/servers/ZoneServer2016/entities/character.ts
```diff
@@ -444,36 +444,55 @@ export class Character2016 extends BaseFullCharacter {
...
+        recipes.push({
+          recipeId: variantRecipeId,
+          itemDefinitionId: recipeDef.ID,
+          nameId: recipeDef.NAME_ID,
+          iconId: recipeDef.IMAGE_SET_ID,
+          unknownDword1: 0, // idk
+          descriptionId: recipeDef.DESCRIPTION_ID,
+          unknownDword2: 1, // idk
+          bundleCount: recipe.bundleCount || 1,
+          membersOnly: false, // could be used for admin-only recipes?
```
QuentinGruber: To answer your comment: it's just a leftover from Planetside. If I'm not mistaken, "members" refers to premium players.

src/servers/ZoneServer2016/managers/craftmanager.ts
QuentinGruber: > Not really, what I do know is that Daybreak used incremental recipe ids (you should be able to find some ids that I found in enums.ts) instead of using the item ids for recipes, maybe if we start using it we might not need it to be an array. But I might be wrong and this is good enough. I'll let @QuentinGruber decide. Maybe he has a better idea

Using recipe ids instead of item ids would simplify both the code and the configuration. If I'm not mistaken, that's already how we handle smelting recipes. I'd say give it a try, and if it ends up being more complicated than expected, we can keep your solution @johnoclockdk

src/servers/ZoneServer2016/entities/character.ts
```diff
@@ -433,9 +433,18 @@ export class Character2016 extends BaseFullCharacter {
...
     let i = 0;
-    for (const recipe of Object.values(this.recipes)) {
-      const recipeDef = server.getItemDefinition(Number(recipeKeys[i]));
-      if (!recipeDef) continue;
+    for (const recipeEntry of Object.values(this.recipes)) {
+      const recipeId = Number(recipeKeys[i]);
+      const recipe = Array.isArray(recipeEntry) ? recipeEntry[0] : recipeEntry;
+
+      // Use rewardId if available, otherwise use the recipe key (recipe ID)
+      const itemId = recipe?.rewardId || recipeId;
```
QuentinGruber: why not making .rewardId mandatory ?

src/servers/ZoneServer2016/managers/craftmanager.ts
```diff
@@ -422,10 +450,63 @@ export class CraftManager {
...
-    if (!recipe) return false;
+
+    // Extract base recipe ID (handle both single and variant recipe IDs)
+    const baseRecipeId = recipeId % 1000000;
+    const variantIndex = Math.floor(recipeId / 1000000);
+
+    const recipeEntry = client.character.recipes[baseRecipeId];
+
+    // Handle recipe arrays - try each variant or specific variant if requested
+    let recipeVariants: any[] = [];
```
QuentinGruber: but you need to still support variant ?  I mean they all have their own recipe now

src/servers/ZoneServer2016/entities/character.ts
johnoclockdk: didnt think of it. but youre right

src/servers/ZoneServer2016/managers/craftmanager.ts
johnoclockdk: you right

johnoclockdk:
@QuentinGruber @QuentinGruber should be ready now thanks for feedback

QuentinGruber (approved):
lgtm, thx @johnoclockdk !

PR #2816, opened 2026-06-20
Fix melee weapon durability on NPC hits

PeepJS:
## Summary
- Apply weapon durability damage in Npc.OnMeleeHit when hitting zombies and other NPCs
- Matches existing behavior for destroyables and crates

### DIFF
--- a/src/servers/ZoneServer2016/entities/npc.ts
+++ b/src/servers/ZoneServer2016/entities/npc.ts
@@ -443,7 +443,7 @@ export class Npc extends BaseFullCharacter {
     const client = server.getClientByCharId(damageInfo.entity),
       weapon = client?.character.getEquippedWeapon();
 
-    if (!client || !weapon) return;
+    if (!weapon) return;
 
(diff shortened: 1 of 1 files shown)


src/servers/ZoneServer2016/entities/npc.ts
```diff
@@ -443,6 +439,17 @@ export class Npc extends BaseFullCharacter {
     if (!this.isAlive) return; // prevent dead npc despawning from melee dmg
     damageInfo.damage = damageInfo.damage / 1.5;
     this.damage(server, damageInfo);
+
+    const client = server.getClientByCharId(damageInfo.entity),
+      weapon = client?.character.getEquippedWeapon();
+
+    if (!client || !weapon) return;
```
QuentinGruber: No need to check both if weapon is client is undefined weapon is too, only !weapon is relevant

QuentinGruber (changes requested):

QuentinGruber (approved):

PR #2822, opened 2026-06-22
Use the right effect for exploder

johnoclockdk:


src/servers/ZoneServer2016/jsms/exploder.jsm.ts
```diff
@@ -154,24 +154,17 @@ export function detonateExploder(
...
-    server,
-    Items.LANDMINE,
-    ownerCharacterId
+  server.sendDataToAllWithSpawnedEntity(
+    server._npcs,
+    npc.characterId,
+    "Command.PlayDialogEffect",
+    {
+      characterId: npc.characterId,
+      effectId: 5242
```
QuentinGruber: We don't have an enum value for this id ?

johnoclockdk: we do

PR #2839, opened 2026-07-06
slotdoesnotcontainitem fix

### LINKED ISSUE #2224: "slotdoesnotcontainitem" while crafting
Player receives message "slotdoesnotcontainitem" in the bottom left when crafting base components from a container and Item disappears.

johnoclockdk:
Root Cause
In craftmanager.ts:618-634, when removing craft components, the code was storing the wrong amount in the removedItems restoration array.

The Problem:
When removeCraftComponent() removes items from a container, it immediately reduces the item's stackCount. The buggy code then stored this modified (reduced) stack count instead of the actual amount removed.

The Fix
I saved the original stack count BEFORE calling removeCraftComponent(), then used that original value when storing in the removedItems array. This ensures that if crafting fails and items need to be restored, the correct amount is restored. The Problem:
When removeCraftComponent() removes items from a container, it immediately reduces the item's stackCount. The buggy code then stored this modified (reduced) stack count instead of the actual amount removed.

### DIFF
--- a/src/servers/ZoneServer2016/managers/craftmanager.ts
+++ b/src/servers/ZoneServer2016/managers/craftmanager.ts
@@ -604,10 +604,11 @@ export class CraftManager {
           }
           removedItems.push({
             itemDS,
-            count: Math.min(itemDS.item.stackCount, remainingItems)
+            count: remainingItems
           });
           remainingItems = 0;
         } else {
+          const originalStackCount = itemDS.item.stackCount;
           if (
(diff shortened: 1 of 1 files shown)


jasonmoonen:
I'll test in a few, need to clear my branch first

QuentinGruber (approved):
Seems good to me but i will let @jasonmoonen test this also.

jasonmoonen (approved):

PR #2841, opened 2026-07-07
screammer nade

johnoclockdk:


johnoclockdk:
there is a issue when holding the greanade it crash the client

src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -3413,6 +3413,72 @@ export class ZoneServer2016 extends EventEmitter {
...
+          }
+        );
+
+        this.pushSound({
+          radius: 100,
+          position: sourceEntity.state.position,
+          agitation: 100
+        });
+
+        // Keep re-emitting the scream noise so AI can stay interested while the
```
johnoclockdk, line 3437: @QuentinGruber We’ll probably need an overwrite behavior for zombies so they keep standing near the grenade. Right now, if a zombie is already chasing someone, it ignores the scream grenade.

johnoclockdk:
ready for review @QuentinGruber

QuentinGruber (approved):
lgtm, gj !

PR #2856, opened 2026-07-29
performance improvement

johnoclockdk:
### Network / Packet Layer

- Movement / vehicle broadcasts: replaced full client scans with the `_entityObservers` index.
- Position / weapon packet serialization: replaced a 13× `Buffer.concat()` chain with a single pre-sized buffer for both protocol versions.
- SOE resend queue: added an early exit after verifying `Map` iteration order remains chronological.
- Packet dispatch:
  - Moved the three hottest packet types to the front of the dispatch switch.
  - Fixed a missing `break` in `SpawnVehicle` that caused `FreeInteractionNpc` to fire on every vehicle spawn.

### Spatial Query Optimizations

- Construction placement & crafting proximity checks → TTL-cached spatial hash.
- Foundation glitch checks (furnace/campfire) → spatial hash (behavior unchanged).
- Crop harvest permission checks → spatial hash (security semantics preserved).
- Zombie, exploder & gasser corpse-smell detection → character spatial grid / `aiTargetSpatialMap`.
- Gasser gas charge & gas cloud damage → character spatial grid / `aiTargetSpatialMap`.
- Screamer scream detection → character spatial grid / `aiTargetSpatialMap`.
- Gas grenade damage ticks → grid narrowed.
- Trap AOE (`SNARE`, `BARBED_WIRE`, `TRAP_FIRE`) → grid narrowed using `getClientsInRange()` / `getVehiclesInRange()`.

### Observer-Based Broadcasts

- NPC death broadcasts → entity observers.
- Crate destruction → entity observers.
- Vehicle destruction → entity observers.
- Shelter door hidden-player reveal → shelter observers.

### Other Optimizations

- Lootable construction destruction → O(1) via existing `mountedCharacter` reference.
- Vehicle spawn collision checks → `.some()` early exit.
- `character.damage()` ping lookup → removed per-hit cross-thread IPC using cached ping.
- Loadout/equipment slot lookups → precomputed `Map`s instead of linear JSON scans.
- `destroyable.pGetLightweight()` → cached rotation quaternion and removed unnecessary position copy.
- `client.searchedProps` → `Array` → `Set` (`indexOf`/`splice`/`includes`/`push` → `has`/`delete`/`add`).
- Added `getDistanceSquared()` and replaced distance calculations where only threshold comparisons were required.

### AI-Assisted PR: This pull request was created with significant AI assistance. Please review it thoroughly and verify all logic, correctness, security, and edge cases before merging.

### DIFF
--- a/src/packets/ClientProtocol/ClientProtocol_1080/shared.ts
+++ b/src/packets/ClientProtocol/ClientProtocol_1080/shared.ts
@@ -637,121 +637,114 @@ export function readPositionUpdateDataAndCheckLength(
 }
 
 export function packPositionUpdateData(obj: any) {
-  let data = Buffer.allocUnsafe(7),
-    flags = 0,
-    v;
-
-  data.writeUInt32LE(obj["sequenceTime"], 2);
-  data.writeUInt8(obj["unknown3_int8"], 6);
+  // Fields are collected into `chunks` and copied into a single pre-sized
+  // buffer at the end instead of repeated Buffer.concat (which recopies
+  // everything written so far on every call) — this is the packer for the
+  // highest-frequency broadcast packet in the game (position updates).
+  let flags = 0;
+  const chunks: Buffer[] = [];
+  let v: Buffer;
 
   if ("stance" in obj) {
     flags |= 1;
-    v = packUnsignedIntWith2bitLengthValue(obj["stance"]);
-    data = Buffer.concat([data, v]);
+    chunks.push(packUnsignedIntWith2bitLengthValue(obj["stance"]));
   }
 
   if ("position" in obj) {
     flags |= 2;
-    v = packSignedIntWith2bitLengthValue(obj["position"][0] * 100);
-    data = Buffer.concat([data, v]);
-    v = packSignedIntWith2bitLengthValue(obj["position"][1] * 100);
-    data = Buffer.concat([data, v]);
-    v = packSignedIntWith2bitLengthValue(obj["position"][2] * 100);
-    data = Buffer.concat([data, v]);
+    chunks.push(packSignedIntWith2bitLengthValue(obj["position"][0] * 100));
+    chunks.push(packSignedIntWith2bitLengthValue(obj["position"][1] * 100));
+    chunks.push(packSignedIntWith2bitLengthValue(obj["position"][2] * 100));
   }
 
   if ("orientation" in obj) {
     flags |= 0x20;
     v = Buffer.allocUnsafe(4);
     v.writeFloatLE(obj["orientation"], 0);
-    data = Buffer.concat([data, v]);
+    chunks.push(v);
   }
 
   if ("frontTilt" in obj) {
     flags |= 0x40;
-    v = packSignedIntWith2bitLengthValue(obj["frontTilt"] * 100);
-    data = Buffer.concat([data, v]);
+    chunks.push(packSignedIntWith2bitLengthValue(obj["frontTilt"] * 100));
   }
 
   if ("sideTilt" in obj) {
     flags |= 0x80;
-    v = packSignedIntWith2bitLengthValue(obj["sideTilt"] * 100);
-    data = Buffer.concat([data, v]);
+    chunks.push(packSignedIntWith2bitLengthValue(obj["sideTilt"] * 100));
   }
 
   if ("angleChange" in obj) {
     flags |= 4;
-    v = packSignedIntWith2bitLengthValue(obj["angleChange"] * 100);
-    data = Buffer.concat([data, v]);
+    chunks.push(packSignedIntWith2bitLengthValue(obj["angleChange"] * 100));
   }
 
   if ("verticalSpeed" in obj) {
     flags |= 8;
-    v = packSignedIntWith2bitLengthValue(obj["verticalSpeed"] * 100);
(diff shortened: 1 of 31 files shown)


src/servers/ZoneServer2016/zoneserver.ts
```diff
@@ -1479,6 +1479,16 @@ export class ZoneServer2016 extends EventEmitter {
         `Character (${characterId}) connection rejected due to local ban`
       );
 
+      for (const c in this._clients) {
+        const banclient = this._clients[c];
+
+        this.sendData(banclient, "H1emu.PrintToConsole", {
```
johnoclockdk: this need removing i was testing stuff my own server

src/packets/ClientProtocol/ClientProtocol_1080/shared.ts
```diff
@@ -637,121 +637,114 @@ export function readPositionUpdateDataAndCheckLength(
...
+  const data = Buffer.allocUnsafe(7 + bodyLength);
   data.writeUInt16LE(flags, 0);
+  data.writeUInt32LE(obj["sequenceTime"], 2);
+  data.writeUInt8(obj["unknown3_int8"], 6);
+
+  let offset = 7;
+  for (const c of chunks) {
+    c.copy(data, offset);
+    offset += c.length;
+  }
```
QuentinGruber, line 747: Buffer.concat does the same internally: iter to get all buffers len , alloc buffer and copy. So idk if there is a win there vs readability.

src/packets/ClientProtocol/ClientProtocol_1080/weapon.ts
```diff
@@ -737,12 +736,22 @@ export function packRemoteWeaponUpdatePacket(obj: any): Buffer {
...
+  let offset = 0;
+  header.copy(data, offset);
+  offset += header.length;
+  transientId.copy(data, offset);
+  offset += transientId.length;
+  updateData.copy(data, offset);
+  offset += updateData.length;
+  packetData.copy(data, offset);
+  return data;
 }
```
QuentinGruber, line 755: Same

src/servers/ZoneServer2016/classes/zoneclient.ts
```diff
@@ -75,7 +75,7 @@ export class ZoneClient2016 {
     ConstructionParentEntity | ConstructionChildEntity
   > = new Set();
   sentInteractionCounter: number = 1;
-  searchedProps: (LootableProp | Lootbag)[] = [];
+  searchedProps: Set<LootableProp | Lootbag> = new Set();
```
QuentinGruber, line 78: good

src/servers/ZoneServer2016/entities/crate.ts
```diff
@@ -176,9 +176,11 @@ export class Crate extends BaseSimpleNpc {
       }
     );
 
-    for (const a in server._clients) {
-      const client = server._clients[a];
-      client.spawnedEntities.delete(server._crates[this.characterId]);
+    const observers = server._entityObservers.get(this.characterId);
+    if (observers) {
+      for (const client of Array.from(observers)) {
```
QuentinGruber, line 181: why array.from ?

QuentinGruber (changes requested):
So i've looked at it a bit.
There is a bunch of optimisation we can do on the packet layer and somes in that PR are probably good.
Same for spatial hash and observers.
But i have to say that i'm against this type of PR.
It's a huge AI PR about optimisation without any benchmark.

Split it, it will be easier to review/test and merge.

You can easily do 4 PRs based on this one.

And do some benchmark, even some simple console.time-trust me-benchmark or if you want an AI generated one like this one https://github.com/QuentinGruber/h1z1-server/blob/master/benchmarks/src/npcAiStress/npc-ai-stress.ts .

johnoclockdk:
@QuentinGruber ill Split this pr up and close it after

johnoclockdk:
> So i've looked at it a bit. There is a bunch of optimisation we can do on the packet layer and somes in that PR are probably good. Same for spatial hash and observers. But i have to say that i'm against this type of PR. It's a huge AI PR about optimisation without any benchmark.
> 
> Split it, it will be easier to review/test and merge.
> 
> You can easily do 4 PRs based on this one.
> 
> And do some benchmark, even some simple console.time-trust me-benchmark or if you want an AI generated one like this one https://github.com/QuentinGruber/h1z1-server/blob/master/benchmarks/src/npcAiStress/npc-ai-stress.ts .

I did run npm run test and the AI tests before pushing. I also told Json that I wasn't sure if I should push it, but he said, Why not? We can always test it and adjust it if anything breaks.

Like I said in my other comment, I'm happy to split it into four separate PRs. I also tested every code change in-game to make sure I didn't break anything. I spent at least two hours testing it in-game, and it's also running on my server (World ID 72). So far, it seems stable.

src/servers/ZoneServer2016/entities/crate.ts
johnoclockdk: Snapshot before mutating spawnedEntities.delete() removes the client from this same observers Set (via TrackedEntitySet's reverse-index), so iterating live would mean deleting out from under the loop. Array.from avoids relying on Set-iterator-safe-deletion semantics.

PR #2861, opened 2026-07-31
Network optimization

johnoclockdk:
Network / Packet Layer

- zoneserver.ts — sendRawToAllOthersWithSpawnedCharacter / sendRawToAllOthersWithSpawnedEntity narrowed to _entityObservers instead of scanning every client.
- soeserver.ts — SOE resend queue: early exit once Map iteration order is confirmed chronological.
- zonepackethandlers.ts — moved the 3 hottest packet types to the front of the dispatch switch; fixed a missing break after Command.SpawnVehicle that was falling through into Command.FreeInteractionNpc on every vehicle spawn.

### DIFF
--- a/src/packets/ClientProtocol/ClientProtocol_1080/shared.ts
+++ b/src/packets/ClientProtocol/ClientProtocol_1080/shared.ts
@@ -637,123 +637,106 @@ export function readPositionUpdateDataAndCheckLength(
 }
 
 export function packPositionUpdateData(obj: any) {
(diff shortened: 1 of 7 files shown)


johnoclockdk:
I left out `ClientProtocol_860/shared.ts` and `weapon.ts`, as well as `ClientProtocol_1080/shared.ts` and `weapon.ts`, because it seemed like you didn't like those changes, @QuentinGruber.

QuentinGruber (approved):

PR #2867, opened 2026-08-07
fix: validate emote ownership on keybind-triggered emotes

FelipeDaros:
## Summary
- `Command.PlayDialogEffect` (sent by the client when an emote keybind is pressed) only validated that the player had fists/hands equipped, without checking whether the player actually owns the emote.
- Non-default emotes (e.g. granted from cases) could be triggered with any `effectId` by a modified client, bypassing the ownership check already enforced by the `/emote` chat command.
- This mirrors the same account-item ownership check used by `/emote` in `handlers/commands/commands.ts`, applying it to the keybind path as well.

---

### DIFF
--- a/src/servers/ZoneServer2016/zonepackethandlers.ts
+++ b/src/servers/ZoneServer2016/zonepackethandlers.ts
@@ -16,6 +16,7 @@
 import { scheduler } from "node:timers/promises";
 import { ZoneClient2016 as Client } from "./classes/zoneclient";
 import { ZoneServer2016 } from "./zoneserver";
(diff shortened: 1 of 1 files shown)


johnoclockdk:
This is a no from me for the following reasons:

1. This was implemented using AI. Please disclose AI usage in the next PR.
2. It doesn’t look like this has been properly tested, as the emote keybind is still not being sent to the server.

Z1Meme:
Closing this pull request, no packet is currently being sent from the client when hitting emote keybinds, which is the core issue. This PR is just adding server logic for clientside functionality that has not yet been unlocked.

PR #2868, opened 2026-08-07
feat: cycle crossbow projectile type with the fire mode switch keybind

FelipeDaros:
## Summary
- The crossbow's weapon definition (`weaponDefinitionId` 1414) has 3 fire groups, one per projectile type: Wooden Arrow, Flaming Arrow, Explosive Arrow.
- Pressing the switch-fire-mode keybind (default V) sends `Weapon.SwitchFireModeRequest`, but the handler only rebroadcast a cosmetic packet to other players and never changed which ammo the weapon actually consumed/reloaded with.
- Added a `firegroupIndex` to the `Weapon` class, and use it in `getWeaponAmmoId` / `getWeaponReloadTime` (previously hardcoded to fire group 0). `Weapon.SwitchFireModeRequest` now cycles through the weapon's fire groups for weapons that expose more than one (crossbow), resets the loaded ammo count (so the player has to reload with the newly selected arrow type), and syncs the change to the shooter and to other players.
- Weapons with a single fire group keep the previous behavior untouched, including the existing ADS-sound workaround.

---

### DIFF
--- a/src/servers/ZoneServer2016/classes/weapon.ts
+++ b/src/servers/ZoneServer2016/classes/weapon.ts
@@ -32,6 +32,9 @@ export class Weapon {
   /** Required for the reload packet to work every time (especially shotgun) */
   currentReloadCount = 0;
 
+  /** Index into the weapon definition's FIRE_GROUPS, selected via Weapon.SwitchFireModeRequest (e.g. crossbow projectile types) */
+  firegroupIndex = 0;
+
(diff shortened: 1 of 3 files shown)


johnoclockdk:
Same as my last comment, this is a no.

1. This was implemented using AI. Please disclose AI usage in the next PR.
2. It doesn’t look like this has been properly tested, as the keybind for swapping ammo types is still not being sent to the server.

PR #2869, opened 2026-08-07
fix: clamp NPC turn rate while attacking so they don't snap instantly

### LINKED ISSUE #2864: Zombies turn too fast
just a simple one if zombies stand and target a player on top of a car or something then they start to turn but the speed is insane allmost instant 180 and amkes it hard to hit them

FelipeDaros:
## Summary
- `Npc.lookAt()` set the NPC's yaw directly to the angle toward its target with no rate limit. When an NPC gets stuck in the Attacking state against a target it can't actually reach (e.g. a player standing on top of a car), it recomputes and snaps its facing on every AI tick, which reads as an almost instant 180-degree turn.
- The same `lookAt()` call from the Attacking state exists identically in 5 melee AIs (zombie, bear, screamer, gasser, wolf), so this wasn't zombie-specific.
- `lookAt()` now accepts an optional `dt` and `maxTurnRateRadPerSec` (default 180deg/s), and clamps the per-call angle change accordingly. All 5 Attacking states now pass their tick's `dt` through. Calls that don't pass `dt` keep the previous instant-snap behavior, so nothing else is affected.

## Test plan
- [x] `npm run lint-quiet`
- [x] `npm run check_prettier`
- [x] `npx tsc -p ./tsconfig.json --noEmit`
- [x] `npm test` (74 tests, 69 passed, 5 skipped requiring Mongo, 0 failed)

---

### DIFF
--- a/src/servers/ZoneServer2016/entities/npc.ts
+++ b/src/servers/ZoneServer2016/entities/npc.ts
@@ -565,15 +565,29 @@ export abstract class Npc extends BaseFullCharacter {
     );
   }
 
-  lookAt(targetPosition: Float32Array) {
+  /**
+   * Turns the NPC to face a target, clamped to maxTurnRateRadPerSec so a
+   * stationary NPC (e.g. attacking a player it can't reach, like one on top
+   * of a car) doesn't snap its facing instantly every AI tick.
+   */
+  lookAt(
+    targetPosition: Float32Array,
+    dt: number = 0,
+    maxTurnRateRadPerSec: number = Math.PI
+  ) {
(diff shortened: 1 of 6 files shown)


johnoclockdk:
this this tested @FelipeDaros ?

QuentinGruber (commented):
It probably work but you still get damages even if the npc doesn't look at you ?

QuentinGruber (changes requested):
Hey @FelipeDaros it works but since the zombies still hit you're in their backs i find it a bit missleading. For me this behavior should be fixed in this PR too.

PR #2871, opened 2026-08-16
Misc client/server packet parity

Z1Meme:


src/packets/ClientProtocol/ClientProtocol_1080/clientUpdate.ts
```diff
@@ -166,35 +166,18 @@ export const clientUpdatePackets: PacketStructures = [
       ]
     }
   ],
-  [
-    "ClientUpdate.ZonePopulation",
```
QuentinGruber, line 170: it was removed on this version ?

src/packets/ClientProtocol/ClientProtocol_1080/clientUpdate.ts
```diff
@@ -166,35 +166,18 @@ export const clientUpdatePackets: PacketStructures = [
...
-      ]
-    }
-  ],
   [
     // 2016
     "ClientUpdate.RespawnLocations",
     0x111f00,
     {
       fields: [
-        { name: "unknownFlags", type: "uint8", defaultValue: 0 },
```
QuentinGruber, line 189: why was it working before ?

QuentinGruber (approved):

src/packets/ClientProtocol/ClientProtocol_1080/clientUpdate.ts
Z1Meme: Yes, 0x112f00 is currently being used twice in this file, 
    "ClientUpdate.StartTimer",
    0x112f00,

and

"ClientUpdate.ZonePopulation",
    0x112f00,


StartTimer is the correct entry.

src/packets/ClientProtocol/ClientProtocol_1080/clientUpdate.ts
Z1Meme: It wasn't, this packet doesn't appear to be sent by the server. ClientUpdateUpdateLockoutTimes is used for showing the red areas on the map when respawning.

PR #2875, opened 2026-08-18
banshee range / block scream under loc of npc

### LINKED ISSUE #2865: banshees on roof tops
banshees on roof tops get triggered by people on the ground, also their triggering distance is a bit high should be a bit lower

johnoclockdk:


johnoclockdk:
this not probbly not the best fix maybe a raycast would be better for this one

QuentinGruber (dismissed):

QuentinGruber (approved):

PR #2886, opened 2026-08-23
fix(construction): align PlacementFinalizeRequest rotation to the 2016 client wire

Z1Meme:
The 0xca0300 schema tail (after BuildingSlot) read a u8 where the client writes a u32 and inserted a phantom float before position2. The two size errors cancel over the 73-byte tail so position2 landed correctly by coincidence, but rotation1/2/3 were parsed 4 bytes early - so the placement yaw (atan2(-w, y) over rotation1) used misaligned quat components and placed structures at the wrong rotation.

Correct the tail to the client wire: unknownDword1 u32, unknownFloat1 f32, unknownByte1 u8, rotation1/2/3 fv4, position2 fv4 (phantom float removed). The handler (ConstructionPlacementFinalizeRequest) already reads rotation1/2/3/position2/itemDefinitionId by name - unchanged; the yaw math is untouched and now reads aligned rotation. Regenerated packet types/interfaces.

### DIFF
--- a/src/packets/ClientProtocol/ClientProtocol_1080/construction.ts
+++ b/src/packets/ClientProtocol/ClientProtocol_1080/construction.ts
@@ -44,12 +44,25 @@ export const constructionPackets: PacketStructures = [
           defaultValue: ""
         },
         { name: "BuildingSlot", type: "string", defaultValue: "" },
-        { name: "unkByte1", type: "uint8", defaultValue: 0 },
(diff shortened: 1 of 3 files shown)


Z1Meme:
This is tested and working in-game, there's no functional change. It's just more accurate when compared to what the client actually sends, verified against the client's actual send method.

QuentinGruber (approved):

johnoclockdk (approved):

PR #2899, opened 2026-09-02
Add nav obstacles for ground tamper and campfire

### LINKED ISSUE #2893: zombies can enter tamper foundations and also shelters and attack players
Image

johnoclockdk:
this should fix #2893


QuentinGruber (approved):

illudedPerception:
i see that the deck foundation is all ready on the list as i made the ticket here i got more infot that they can also pop through decks if low enough 
image

